PR #63, opened 2015-02-13
General fixes for Fixes

dcrissman:
Changes made by @derek63 and myself, primarily to address changes to the migrator entities.


consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -18,382 +18,393 @@
...
-	protected Map<String, JsonNode> findDestinationData(LightblueRequest findRequest) throws IOException {
-		return getJsonNodeMap(getDestinationClient().data(findRequest, JsonNode[].class), getJobConfiguration().getDestinationEntityKeyFields());
-	}
-
-	protected LinkedHashMap<String, JsonNode> getJsonNodeMap(JsonNode[] results, List<String> entityKeyFields) {
-		LinkedHashMap<String, JsonNode> resultsMap = new LinkedHashMap<>();
-		for(JsonNode result : results) {
-			StringBuilder resultKey = new StringBuilder();
-			for(String keyField : entityKeyFields) {
-				resultKey.append(result.findValue(keyField)).append("|||");
```
jewzaam: The delimiter here was to ensure value fields don't incorrectly collide.  For example, if you have numeric identify fields it's very possible for identity field 123456 to collide with an object that has two identity fields: 123 and 456.

jewzaam:
Review complete, just the one comment about the delimiter for the map key.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
dcrissman: @derek63 would you mind taking a look at this?

jewzaam: @dcrissman this is your PR.  Would you rather we merge this PR and deal with the delimiter in another PR?

derek63: Delimiter makes sense for the multiple-field identities, just need to add logic to parse/remove it  when building the query to lightblue based on identity fields (as we were submitting queries with it in there)

dcrissman: I can no longer reproduce so I  reverted that particular commit. If it comes up again, we can tackle it in another issue.

PR #64, opened 2015-02-16
Handle NullNode case

dcrissman:
This change will help the consistency checker better handle NullNodes.

Also handle the case when ComparisonExclusionPaths is null.

Added supporting unit tests.

### DIFF
--- a/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
+++ b/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
@@ -379,8 +379,8 @@ public class MigrationJob implements Runnable {
                 JsonNode sourceNode = sourceEntry.getValue();
                 JsonNode destinationNode = destinationDocument.findValue(sourceEntry.getKey());
 
-                if((sourceNode == null || JsonNodeType.NULL.equals(sourceNode.getNodeType())
-                        && (destinationNode == null || JsonNodeType.NULL.equals(destinationNode.getNodeType())))) {
+                if((sourceNode == null || JsonNodeType.NULL.equals(sourceNode.getNodeType()))
+                        && (destinationNode == null || JsonNodeType.NULL.equals(destinationNode.getNodeType()))) {
                     continue;
                 }
                 if(!sourceNode.equals(destinationNode)) {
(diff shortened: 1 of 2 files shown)


consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -371,9 +372,18 @@ protected boolean documentsConsistent(JsonNode sourceDocument, JsonNode destinat
...
-                if(!sourceNode.getValue().equals(destinationDocument.findValue(sourceNode.getKey()))) {
+            Entry<String, JsonNode> sourceEntry = nodeIterator.next();
+
+            List<String> excludes = getJobConfiguration().getComparisonExclusionPaths();
+            if(excludes == null || !excludes.contains(sourceEntry.getKey())) {
+                JsonNode sourceNode = sourceEntry.getValue();
+                JsonNode destinationNode = destinationDocument.findValue(sourceEntry.getKey());
+
+                if((JsonNodeType.NULL.equals(sourceNode.getNodeType())
+                        && destinationNode == null)){
```
jewzaam: Both `sourceNode` and `destinationNode` are instances of `JsonNode` but the null check conditions are different.  Would this be better?

```java
if((sourceNode == null || JsonNodeType.NULL.equals(sourceNode.getNodeType())
      && (destinationNode == null || JsonNodeType.NULL.equals(destinationNode.getNodeType()))) {
```

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -371,9 +372,18 @@ protected boolean documentsConsistent(JsonNode sourceDocument, JsonNode destinat
...
+                    continue;
+                }
+                if(!sourceNode.equals(destinationNode)) {
```
jewzaam: Seems safer to check that `sourceNode` isn't null, but not sure that it could happen.  Anyway, can't hurt to check for null

dcrissman: In my testing sourceNode was a NullNode and destination was simply null, which is why I wrote it that way. But I like your suggestion as it is a more complete check. I will make changes.

PR #68, opened 2015-02-18
Fixes #65 - Consistent check handle array and object nodes

### LINKED ISSUE #65: MigrationJob does not appear to iterate over items in sub-arrays for consistency comparisons
The migrator seems to be attempting to compare entire arrays, instead of each individual element. I suspect we would want ways to exclude elements in sub-arrays/objects just like top level elements.

The other question this brings to mind is does it handle sub-objects in the same way?

-
source:
> [{"statusCode":"active","termsVerbiageTranslation":[{"statusCode":"active","localeCodeText":"en_US","publishedFlag":true,"url":" "}],"versionText":"1","description":"Evaluation Terms"}]

-
destination:
> [{"statusCode":"active","uid":"bd70bd99-b443-4ce2-bd02-f87c772a6513","termsVerbiageTranslation":[{"uid":"c65d6f5e-ffdd-4bba-940c-fea307790bfa","localeCodeText":"en_US","publishedFlag":true,"url":" ","statusCode":"active","creationDate":"20150216T18:07:27.880+0000"}],"versionText":"1","description":"Evaluation Terms","name":"someName","termsVerbiageTranslation#":1}]

dcrissman:
* documentsConsistent has been changed to being a recursive method that will iterate over children nodes
* comparisonExclusionPaths now supports notation similar to Paths, allowing you to specify children nodes to ignore
* switched from using findValue() to get() on destination node. This is because findValue() searches children in addition to the parent, causing undesired results when the same field name is used in multiple places.

### DIFF
--- a/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
+++ b/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
@@ -15,11 +15,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.redhat.lightblue.client.LightblueClient;
 import com.redhat.lightblue.client.enums.SortDirection;
 import com.redhat.lightblue.client.expression.query.Query;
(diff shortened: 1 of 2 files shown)


dcrissman:
That unit test can be ignored until we are doing size checks on source and destination objects.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -367,29 +370,77 @@ private int overwriteLightblue(List<JsonNode> documentsToOverwrite) {
...
     protected boolean documentsConsistent(JsonNode sourceDocument, JsonNode destinationDocument) {
-        boolean consistent = true;
+        return doDocumentsConsistent(sourceDocument, destinationDocument, null);
+    }
 
-        Iterator<Entry<String, JsonNode>> nodeIterator = sourceDocument.fields();
+    //Recursive method
+    private boolean doDocumentsConsistent(final JsonNode sourceDocument, final JsonNode destinationDocument, final String path) {
+        List<String> excludes = getJobConfiguration().getComparisonExclusionPaths();
+        if(excludes != null && excludes.contains(path)) {
```
jewzaam: Check that `path is not null.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -367,29 +370,77 @@ private int overwriteLightblue(List<JsonNode> documentsToOverwrite) {
...
+            ArrayNode sourceArray = (ArrayNode) sourceDocument;
+            ArrayNode destinationArray = (ArrayNode) destinationDocument;
+
+            if(sourceArray.size() != destinationArray.size()){
+                return false;
+            }
+
+            //assumed positions in the array should be the same, else inconsistent.
+            for(int x = 0; x < sourceArray.size(); x++){
+                if(!doDocumentsConsistent(sourceArray.get(x), destinationArray.get(x), path)){
```
jewzaam: Shouldn't `path` be updated to be path + "." + array field name? (or just array field name in case of top level field)

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
dcrissman: For the first run, the path will be null by design. I am handling the null case further down on line 432. StringUtils.isEmpty(path) will check for null or an empty string.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
dcrissman: Not for arrays, no. Take the following path for example, in it both termsVerbiage and termsVerbiageTranslation are technically arrays, but we don't care for excluding purposes as the intention that all elements in an array would ignore the same fields. In this regard, it probably differs from the Path notation used in lightblue proper.

termsVerbiage.termsVerbiageTranslation.publishedFlag

jewzaam: Example:

```json
{
   "array1": [
      {
         "field1": "value1"
      }
   ]
}
```
When passing each child of `array1` the path should be `"array1"` and not `null`.

jewzaam: What is path then?  It's not really path.  I think what you're saying is the exclusion paths are anything to exclude without any array context.  What if my entity has "createdDate" at the root and in an array, but I only want to exclude from the array?  I have to provide context by saying the name of the array.

dcrissman: My assumption is that an array would never be top level for our purposes, but maybe that is a bad assumption? If it is not a top level, but rather wrapped in a object node, then the Object logic would ensure that the array's field name is present in the path.

jewzaam: I think this needs a unit test, where document looks like this:

```json
{
  "field1": "value1",
  "something": "x",
  "array": [
    {
      "field2": "value2",
      "something": "y"
    }
  ]
}
```

And excluded field is only "array.something"

jewzaam: Yea, bad assumption to expect no top level array fields.

dcrissman: The unit test you requested has been added.

I will need to look into top level arrays then.

dcrissman: Added additional unit tests for top level arrays.

consistency-checker/src/test/java/com/redhat/lightblue/migrator/consistency/MigrationJobTest.java
```diff
@@ -139,6 +216,109 @@ public void testDocumentsConsistent_SourceAndDestinationNotConsistent(){
...
+
+        assertTrue(migrationJob.documentsConsistent(sourceNode, destNode));
+    }
+
+    /**
+     * Ensures that a more complex json document, with a deep invalid field but where that field is excluded.
+     * Should pass.
+     */
+    @Test
+    public void testDocumentsConsistent_ComplexJson_Pass() throws Exception{
```
jewzaam: A few things to make round out the tests, add some that:
* have a different value for "something" at the top level, verify docs are not consistent
* exclude "something" instead of "array.something" and verify documents are not consistent

dcrissman: Tests added.

consistency-checker/src/test/java/com/redhat/lightblue/migrator/consistency/MigrationJobTest.java
```diff
@@ -139,6 +216,147 @@ public void testDocumentsConsistent_SourceAndDestinationNotConsistent(){
...
+
+        assertFalse(migrationJob.documentsConsistent(source, dest));
+    }
+
+    /**
+     * Have a different value for "something" at the top level, verify docs are not consistent
+     * Should fail.
+     */
+    @Test
+    public void testDocumentsConsistent_ComplexJson_WithTopLevelMismatch_fail() throws Exception{
```
jewzaam: I meant these, if they're not there somewhere else (sorry, didn't dig to find out):

```java
        ObjectMapper mapper = new ObjectMapper();

        JsonNode source = mapper.readTree(
                "[{\"field1\": \"value1\", \"something\": \"x\", \"array\": [{\"field2\": \"value2\",\"something\": \"y\"}]}]");

        JsonNode dest = mapper.readTree(
                "[{\"field1\": \"value1\", \"something\": \"y\", \"array\": [{\"field2\": \"value2\",\"something\": \"z\"}]}]");

        MigrationConfiguration jobConfiguration = new MigrationConfiguration();
        jobConfiguration.setComparisonExclusionPaths(Arrays.asList("something"));
        migrationJob.setJobConfiguration(jobConfiguration);

        assertFalse(migrationJob.documentsConsistent(source, dest));
```


```java
        ObjectMapper mapper = new ObjectMapper();

        JsonNode source = mapper.readTree(
                "[{\"field1\": \"value1\", \"something\": \"x\", \"array\": [{\"field2\": \"value2\",\"something\": \"y\"}]}]");

        JsonNode dest = mapper.readTree(
                "[{\"field1\": \"value1\", \"something\": \"y\", \"array\": [{\"field2\": \"value2\",\"something\": \"y\"}]}]");

        MigrationConfiguration jobConfiguration = new MigrationConfiguration();
        jobConfiguration.setComparisonExclusionPaths(Arrays.asList("something"));
        migrationJob.setJobConfiguration(jobConfiguration);

        assertTrue(migrationJob.documentsConsistent(source, dest));
```

dcrissman: I think everything is tested at a low level, but it is good to have some complex json tests also.
Changes made.

PR #69, opened 2015-02-19
Added script to generate lightblue insert request for creating migration jobs

derek63:


jewzaam:
chmod +x ./consistency-checker/migrationJobs.sh

consistency-checker/migrationJobs.sh
```diff
@@ -0,0 +1,50 @@
...
+START_DATE=$2
+END_DATE=$3
+NUM_DAYS=$4
+EXPECTED_EXECUTION_TIME=$5
+CREATED_BY=$6
+AVAILABLE_DATE=$7
+
+if [ $1"x" == "x" ] || [ $2"x" == "x" ] || [ $3"x" == "x" ] || [ $4"x" == "x" ] || [ $5"x" == "x" ] || [ $6"x" == "x" ]; then
+    echo "Usage: ./migrationJobs.sh <entityName> <startDate> <endDate> <frequency> <expectedExecutionTime> <createdBy> <availableDate>"
+    echo "Example: ./migrationsJobs.sh user 2014-01-01 2014-12-31 1 30000 derek63 2015-03-01"
```
jewzaam: /migrationsJobs/migrationJobs/

jewzaam:
How is this script used?  Would be better to either call the service to create the records or to have it generate the insert request.  Right now the output is a bunch of individual json documents with no clear delimiter between them.  Simplest route is generate the full request body and leave it to the user to call the service.

derek63:
"Simplest route is generate the full request body and leave it to the user to call the service."  That is actually what this does currently.  The script outputs a JSON document that is a lightblue insert request that contains all the migration jobs as part of the data section.

derek63:
Just realized I was missing a comma between the migrationJob documents in the data section and pushed a change for that.

consistency-checker/migrationJobs.sh
derek63: Pushed a change to fix the example script name

consistency-checker/migrationJobs.sh
```diff
@@ -0,0 +1,51 @@
...
+    echo  \"lastUpdatedBy\": \"$CREATED_BY\"
+    echo }
+    current="$next"
+    [ "$next" \< "$END_DATE" ] || break
+    let i=i+1
+    echo ,
+done
+
+echo  ],
+echo  \"projection\": [
```
jewzaam: Is there a reason to have the projection?  If you're bulk loading a ton of objects it isn't likely you need to see the details of the object, just the count that were added.  I guess if you don't want them it could be manually edited out so not a big deal.

jewzaam:
Date formats do not match documented[1][2] format: YYYYMMDD HHmmssSSSS+-ZZZZ

[1] http://dev.docs.lightblue.io/key_decisions/README.html
[2] http://jewzaam.gitbooks.io/draft-lightblue-specifications/content/backend_specifications/mongodb/metadata.html

consistency-checker/migrationJobs.sh
derek63: I thought that the insert request required that a projection is defined, which is why it is included here, however I just tried without one, and it seemed to work fine.  So I guess we could remove this altogether, or modify it to just return the _id field so we could could confirm all the ones in the request got added.

derek63:
Running the script generates a date in the following format, 

"whenAvailableDate" : "20150221T00:00:00.000+0000"

which does seem to match what is returned when I query already defined internal entities with date fields

"creationDate" : "20141203T17:59:07.168+0000"

These are the same format, right?

consistency-checker/migrationJobs.sh
derek63: Just pushed a commit to only return the _id of documents inserted

jewzaam:
[Format in code](https://github.com/lightblue-platform/lightblue-core/blob/127902a8f187c8a6191d0bf0d7cf72b823092e60/metadata/src/main/java/com/redhat/lightblue/metadata/types/DateType.java#L41) matches what you're using.  I'll add issues to fix the documentation.

PR #71, opened 2015-02-23
use correct daemon version

dcrissman:
Migrator is currently building with a newer version of the client then what is on the server. Simply making the version number match.


dcrissman:
Closing for now

PR #73, opened 2015-02-25
don't attempt to fetch destination documents if there are no source documents

dcrissman:
write a warn message as opposed to an ugly exception


derek63:
This build is apparently never going to start/complete, so I downloaded your branch and verified all looks good locally.  Merging now

derek63:
I guess the branch did actually build successfully, just didn't get reported back here.

PR #75, opened 2015-02-26
Adding shared migration code

paterczm:
DAOFacadeBase and LightblueMigrationPhase (testing helper).


utils/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+    <groupId>com.redhat.lightblue.migrator</groupId>
+    <artifactId>lightblue-migrator-pom</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+  </parent>
+  <groupId>com.redhat.lightblue.migrator</groupId>
+  <artifactId>lightblue-migrator-utils</artifactId>
+  <packaging>jar</packaging>
+  <version>0.0.1-SNAPSHOT</version>
+  <name>lightblue-migrator: ${project.groupId}|${project.artifactId}</name>
+  <dependencies>
```
jewzaam: Define dependencies in root pom's dependencyManagement to move management of all versions except artifacts built by this repo to the root pom.

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
```diff
@@ -0,0 +1,198 @@
...
+ *
+ */
+@SuppressWarnings("all")
+public class DAOFacadeBase {
+
+    private static final Logger log = LoggerFactory.getLogger(LightblueMigrationConfiguration.class);
+
+    protected final Object legacyDAO, lightblueDAO;
+
+    public DAOFacadeBase(Object legacyDAO, Object lightblueDAO) {
```
jewzaam: I'd make this an abstract class and use generics to get away from using Object.  I know it's just using reflection on these objects, but it will allow for some type checking, forcing legacyDAO and lightblueDAO to implement the same interface.

paterczm: I didn't make it abstract on purpose. If you have a DAO which already has a parent, it's problematic to use an abstract class. I like the idea to use generics for additional type safety.

PR #91, opened 2015-03-10
Fixes #84: update to have one executor process a job

### LINKED ISSUE #84: Multiple migrators running the same set of jobs will duplicate work
This is because a migrator will greedy fetch all jobs and keep them in memory. So even after a job is updated by another migrator, other instance would never get that information as it has already essentially been cached.

jewzaam:
If multiple migrator instances try to process a job the execution that was put in the array first wins.  All other executions are marked as complete and end date is set.  The indication that nothing was done is in the document counts.  There is no special flag or anything to indicate it didn't do any work other than those counts.

### DIFF
--- a/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
+++ b/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
@@ -7,7 +7,6 @@ import static com.redhat.lightblue.client.expression.query.ValueQuery.withValue;
 import static com.redhat.lightblue.client.projection.FieldProjection.includeFieldRecursively;
 
 import java.io.IOException;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -33,7 +32,6 @@ import com.redhat.lightblue.client.enums.SortDirection;
 import com.redhat.lightblue.client.expression.query.Query;
 import com.redhat.lightblue.client.expression.query.ValueQuery;
 import com.redhat.lightblue.client.expression.update.AppendUpdate;
-import com.redhat.lightblue.client.expression.update.ForeachUpdate;
 import com.redhat.lightblue.client.expression.update.ObjectRValue;
 import com.redhat.lightblue.client.expression.update.PathValuePair;
 import com.redhat.lightblue.client.expression.update.SetUpdate;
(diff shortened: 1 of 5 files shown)


consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -260,38 +268,80 @@ public void run() {
...
+                currentRun.setActualEndDate(new Date());
+                saveJobDetails(jobExecutionPsn);
+            } else {
+                // just mark complete and set actual end date
+                markExecutionComplete(jobExecutionPsn);
+            }
+
+        } catch (RuntimeException e) {
+            // would be nice to reference DataType.DATE_FORMAT_STR in core..
             SimpleDateFormat dateFormat = new SimpleDateFormat("yyyyMMdd'T'HH:mm:ss.SSSZ");
```
derek63: You can use the one that is in lightblue client ClientConstants.getDateFormat()

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -260,38 +268,80 @@ public void run() {
...
+            LightblueResponse response = saveJobDetails(jobExecutionPsn);
 
-            Map<String, JsonNode> destinationDocuments = getDestinationDocuments(sourceDocuments);
+            boolean processJob = true;
 
-            List<JsonNode> documentsToOverwrite = getDocumentsToOverwrite(sourceDocuments, destinationDocuments);
+            // verify this job is the first active thread processing the thread.
+            // first extract processed data as MigrationJob
+            JsonNode node = response.getJson().path("processed");
+            try {
```
derek63: Should we pull out the processJob (between here and line 311) logic into a separate method?

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -260,38 +268,80 @@ public void run() {
...
+                }
 
-            if (!documentsToOverwrite.isEmpty()) {
-                hasInconsistentDocuments = true;
+                // first incomplete job execution must be for this execution else we don't process it
+                int i = 0;
+                for (MigrationJobExecution execution : jobs[0].getJobExecutions()) {
+                    // if we find an execution that is not our pid but is active
+                    // in the array before ours, we do not get to process
+                    if (!execution.isCompletedFlag() && !pid.equals(execution.getPid())) {
```
dcrissman: How are we handling the case where a job has been idle for some period of time? Presumably the process had an error and needs to be attempted again. Maybe the process on the server has been restarted so the pid is no longer the same.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
jewzaam: done

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
jewzaam: done

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
jewzaam: done, hard coded to time out job execution after 60 minutes

PR #95, opened 2015-03-11
Add Hystrix support to migrator

dcrissman:
Note: Hystrix is non-optional at this time.

### DIFF
--- a/consistency-checker/pom.xml
+++ b/consistency-checker/pom.xml
@@ -20,6 +20,10 @@
       <artifactId>lightblue-client-http</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.redhat.lightblue.client</groupId>
+      <artifactId>lightblue-client-hystrix</artifactId>
+    </dependency>
+    <dependency>
       <groupId>commons-cli</groupId>
       <artifactId>commons-cli</artifactId>
     </dependency>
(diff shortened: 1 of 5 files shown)


jewzaam:
Missing configuration.  Default is isolation of THREAD with a timeout of 1000 msec.  We might want to swap to SEMAPHORE for this client and increase concurrent requests from the default of 10.

jewzaam:
https://github.com/Netflix/Hystrix/wiki/Configuration#CommandExecution

dcrissman:
@jewzaam would those be changes for the puppet layer?

jewzaam:
For migrator I would say just bundle it as a resource for now.  We can refactor it out later.  We're not using hystrix to limit concurrent jobs so just set the concurrent requests to something high.  We want hystrix in this case for the ease of getting metrics.

We could put this as part of the client/hystrix module.  Not sure the implications to making changes to configs for clients.  Standalone apps like migrator could pass a command line arg.  For jboss deployments having many clients in a container it gets harder.

dcrissman:
@jewzaam changes made, ready for review

dcrissman:
Still doing some testing... please hold off on review. Thanks.

dcrissman:
@jewzaam This is ready for review.

PR #100, opened 2015-03-13
add debug statements

dcrissman:
Add some debug logging statements.


bserdar:
Instead of Logger.debug("blah:"+yada), use Logger.debug("blah:{}",yada)

dcrissman:
@bserdar done

PR #103, opened 2015-03-17
protects against 0 and negative thread values, either case will be considered a skip

dcrissman:
Thread count was set to 0 in job configuration entity.

### DIFF
--- a/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
+++ b/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
@@ -133,6 +133,10 @@ public class ConsistencyChecker implements Runnable {
             List<MigrationConfiguration> configurations = getJobConfigurations();
 
             for (MigrationConfiguration configuration : configurations) {
+                if (configuration.getThreadCount() < 1) {
+                    continue;
+                }
+
                 configuration.setConfigFilePath(configPath);
                 configuration.setMigrationJobEntityVersion(migrationJobEntityVersion);
                 List<MigrationJob> jobs = getMigrationJobs(configuration);


luan-cestari:
I would recommend that the initial value of that class be a valid value https://github.com/dcrissman/lightblue-migrator/blob/no-threads/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationConfiguration.java#L13 (so it could be have a default valid value and (optionally but I think it make sense specially for DDD ) you could puts that business rule into the set method (and on the constructor if you want to make a new constructor as well) )

dcrissman:
@luan-cestari In this case, the value was intentionally set to a value of 0 in the entity. I updated the description to be more clear.

luan-cestari:
@dcrissman Alright. Could you file an issue on the migrator's documentationto be updated with this information (and other remaining)? Let me know if you want to add javadoc about it or not so  I will merge this PR

jewzaam:
@luan-cestari the metadata for migrationConfiguration shows this is a required field, so it will always be set in the database.  We could supply a default value for the field but it will always be overriden.

https://github.com/lightblue-platform/lightblue-migrator/blob/master/consistency-checker/src/main/resources/migrationConfiguration.json#L97-L103

luan-cestari:
@jewzaam yeah, but I just mentioned on my last comment that it might be good to update the documentation about this change as well ( such as that json snippet you mentioned https://github.com/lightblue-platform/lightblue-migrator/blob/master/consistency-checker/src/main/resources/migrationConfiguration.json#L97-L103 ) . I don't see any other problem than that right now (an it is not a big issue right now)

dcrissman:
@luan-cestari As we do not yet have documentation for the migrator, I mentioned this change in the docs issue to do so.

PR #109, opened 2015-03-18
Updated to log how long consistency checker is waiting for new work

jewzaam:
Default wait was also updated, clarified constant name to indicate it's msec unit

### DIFF
--- a/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
+++ b/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
@@ -33,9 +33,9 @@ public class ConsistencyChecker implements Runnable {
 
     public static final Logger LOGGER = LoggerFactory.getLogger(ConsistencyChecker.class);
 
-    public static final int MAX_JOB_WAIT_TIME_MSEC = 24 * 60 * 60 * 1000; // 24 hours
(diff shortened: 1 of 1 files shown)


consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
```diff
@@ -155,12 +155,16 @@ public void run() {
...
                 if (run && !Thread.interrupted()) {
                     MigrationJob nextJob = getNextAvailableJob();
-                    long timeUntilNextJob = DEFAULT_WAIT;
-                    if (nextJob != null) {
+                    long timeUntilNextJob = DEFAULT_JOB_WAIT_MSEC;
+                    if (nextJob != null && nextJob.getWhenAvailableDate() != null) {
                         timeUntilNextJob = nextJob.getWhenAvailableDate().getTime() - new Date().getTime();
                     }
+                    if (timeUntilNextJob > MAX_JOB_WAIT_MSEC) {
+                        timeUntilNextJob = MAX_JOB_WAIT_MSEC;
```
luan-cestari: I know that it might be documented already about this, but It would be nice to have a log entry here (warn level?) as someone can misuse migrator and get undesired behavior without anything being logged about it. What do you think?

jewzaam:
@luan-cestari added log when falling back to default
Also simplified the constants we use for waiting and reduced times to something more reasonable.

luan-cestari:
@jewzaam awesome. I will merge it as soon as Travis (successfully) build it

PR #110, opened 2015-03-18
Fixes #106: range for fetching next job is (0, 0)

jewzaam:
Also added condition to how jobs are fetched for processing and added some high level logging.

PR #113, opened 2015-03-18
fixes #107 - ignores jobs that are for execution later than today, also added some logging

### LINKED ISSUE #107: getMigrationJobs shouldn't fetch unavailable jobs
Add clause about whenAvailableDate <= current date

dcrissman:


luan-cestari:
Is it possible to put a periodic checker to capture and inform (probably just log) the pending jobs that are not completed and not for today (maybe due some problems or any other issue)? I think it will be important to the user know this

jewzaam:
Migrator only checks for the next job that would be processed, it doesn't fetch the count of outstanding jobs.  This could be queried via crud api's.

@dcrissman this is going to conflict with https://github.com/lightblue-platform/lightblue-migrator/pull/110/ a bit

PR #120, opened 2015-03-20
Fixes #97 - Better status for the migration jobs

### LINKED ISSUE #97: Need better status on job executions
Right now there's just a boolean flag indicating if the job execution is complete (jobExecutions.*.isComplete).  If the source for the job has no documents then there is no way to distinguish this job execution from an execution that was terminated for some other reason (running too long, process died, aborted because of duplicate processing).

I propose removing isComplete and replacing it with a string field 'statusCode' that is backed by an enum.  Values of the enum would be something like:
* running - currently being processed
* abort_duplicate - execution aborted because it was a duplicate
* abort_timeout - execution aborted because of timeout (running too long)
* complete_failed - execution completed but failed
* complete_partial - execution completed but only partially
* complete_success - execution completed successfully

luan-cestari:
More info on  https://github.com/lightblue-platform/lightblue-migrator/issues/97


consistency-checker/src/main/resources/migrationJob.json
```diff
@@ -133,12 +133,12 @@
                             },
                             "description": "Actual end date for this migration job"
                         },
-                        "completedFlag": {
-                            "type": "boolean",
+                        "jobStatus": {
+                            "enum":[ "NEW","READY","STARTING","RUNNING","RUNNING_ASYNC","FINISHING","COMPLETED_SUCCESS","COMPLETED_PARTIAL","COMPLETED_FAILED","ABORTING","ABORTED_DUPLICATE","ABORTED_TIMEOUT","ABORTED_AUTH","ABORTED_UNKNOWN","UNKNOWN" ],
```
jewzaam: Enums are defined in entityInfo.  See [this](https://github.com/lightblue-platform/lightblue-core/blob/master/metadata/src/test/resources/metadata-json-schema-test-valid/schema-test-metadata-object-object.json#L10-L15) for defining an enum and [this](https://github.com/lightblue-platform/lightblue-core/blob/master/metadata/src/test/resources/metadata-json-schema-test-valid/schema-test-metadata-object-object.json#L39-L44) for using the enum.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/JobStatus.java
```diff
@@ -0,0 +1,74 @@
+package com.redhat.lightblue.migrator.consistency;
+
+/**
+ *
+ * JobStatus can be one of the following values:
+ *      NEW                - Means the Job just created and it isn't ready to be executed
+ *      READY              - Means the Job is ready to be executed but not in execution
```
jewzaam: Status is on a job execution record.  This is only created when a job is being executed.  Status of NEW and READY will not be part of execution.  NEW is implied by the migrationJob existing in the db without any elements in jobExecutions.  READY is implied by the migrationJob existing in the db without any elements in jobExecutions and whenAvailableDate <= new Date()

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/JobStatus.java
```diff
@@ -0,0 +1,74 @@
...
+ *      STARTING           - It just got to be processing
+ *      RUNNING            - Currently being processed for synchronous request
+ *      RUNNING_ASYNC      - Currently being processed for asynchronous request
+ *      FINISHING          - The execution is almost complete, just few more steps to move to a final status
```
jewzaam: Having status of FINISHING and ABORTING requires additional writes to the entity.  I don't know that there's sufficient value to track these.  If a job is either finishing or aborting I assume it will very quickly be done.  There shouldn't be any cleanup of data to do, so no external system interactions to slow things down.  It's all in memory, and not sure exactly what would be *done* after these states.

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
```diff
@@ -202,7 +202,9 @@ public void run() {
             findRequest.where(and(
                     withValue("configurationName = " + configuration.getConfigurationName()),
                     withValue("whenAvailableDate <= " + ClientConstants.getDateFormat().format(new Date())),
-                    not(withSubfield("jobExecutions", withValue("completedFlag = true")))));
+                    not(withSubfield("jobExecutions", withValue("jobStatus = 'COMPLETED_SUCCESS'"))),
+                    not(withSubfield("jobExecutions", withValue("jobStatus = 'COMPLETED_PARTIAL'")))
```
jewzaam: The getMigrationJobs function should not return jobs that are currently being executed, but needs to pick up jobs that were marked for execution a long time ago.  So, a job can be returned by this function when:

```
and:
    whenAvailableDate <= new Date()
    not:
        jobExecutions.jobStatus $in ['COMPLETED_SUCCESS', 'COMPLETED_PARTIAL']
    and:
        not: 
            jobExecutions.jobStatus $not_in [COMPLETED_SUCCESS', 'COMPLETED_PARTIAL']
        jobExecutions.actualStartDate < (new Date() - expectedExecutionMilliseconds)
```

but please check my logic here.  We want to pickup work that has been marked in progress for too long.

consistency-checker/src/main/resources/migrationJob.json
luan-cestari: Thanks for the hint. Fixed, I will update the PR

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/JobStatus.java
luan-cestari: I can remove those two status (I think that what you mean) but one think that I thought to at least leave one of them is if we got the JVM crashed or something (example there is a way to record the JVM state and reproduce) and if we got that object we would know that it wasn't running. But I think it is a very uncommon use case, so I will remove NEW and READY

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/JobStatus.java
luan-cestari: OK. Removing FINISHING and ABORTING

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
luan-cestari: I think there is something wrong in the snippet you made (or maybe there is something around lightblue-client that I'm not familiar with). The 'jobExecutions.jobStatus' is mentioned twice with contradictions between each other. I think maybe you would like to put an 'or' instead the second 'and'. I will leave this piece of code as it is right now so we can change when we conclude this thread.

jewzaam: I think you're right, thanks!  Let me go with some scenarios since it's confusing me too:

### Scenarios:
Each scenario's condistions are AND.  Expectation noted for each scenario.

#### future dated
* whenAvailableDate > new Date()

Processable? No.

#### available and never processed
* whenAvailableDate <= new Date()
* jobExecutions is an empty array

Processable? Yes!

#### available and processed
* whenAvailableDate <= new Date()
* jobExecutions at least one jobStatus equal to 'COMPLETED_SUCCESS' or 'COMPLETED_PARTIAL'

Processable? No.

#### available and currently processing
* whenAvailableDate <= new Date()
* jobExecutions is not empty
* jobExecution has no jobStatus equal to 'COMPLETED_SUCCESS' or 'COMPLETED_PARTIAL'
* max(jobExecutions.actualStartDate) >= new Date() - expectedExecutionMilliseconds)

Processable? Yes!

#### available, currently processing, but presumed "dead"
* whenAvailableDate <= new Date()
* jobExecutions is not empty
* jobExecution has no jobStatus equal to 'COMPLETED_SUCCESS' or 'COMPLETED_PARTIAL'
* max(jobExecutions.actualStartDate) < new Date() - expectedExecutionMilliseconds)

Processable? No.

### Query:
Therefore the query would be:
```
and:
    whenAvailableDate <= new Date()
    or:
        jobExecutions# = 0
        not:
            jobExecutions.jobStatus $in ['COMPLETED_SUCCESS', 'COMPLETED_PARTIAL']
        and:
            not: 
                jobExecutions.jobStatus $not_in [COMPLETED_SUCCESS', 'COMPLETED_PARTIAL']
            jobExecutions.actualStartDate < (new Date() - expectedExecutionMilliseconds)
```

luan-cestari: I think the query is right now by the axioms you made.  Maybe we should also ignore the MigrationJobExecution that were aborted.

I also have a doubt if jobExecutions# would be greater than 1 or, if it is greater, if the reaming items would be processed as the run() method would called just once by the ConsistencyChecker on https://github.com/luan-cestari/lightblue-migrator/blob/8955b9c69e23d4cf17cc3187ed700d497ee2f87c/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java#L153-153 . In other words, my concern is if all MigrationJobExecution from a MigrationJob are being executed as expected (which I guess not) and if we are processing all MigrationJobExecution , would this query also include the MigrationJobs that at least one of MigrationJobExecution.jobStatus is not Completed ?

jewzaam: Only one execution should ever be processed at a time.  But that's something that has to be handled somewhere else because of how updates to the migrationJob entity are done.  This query is just to get a collection of what is processable based on what is in the db at the time the query was executed.

luan-cestari: OK. I will update the query in this PR. I will also introduce the changes to change the status we previously discussed

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -281,15 +287,83 @@ public void run() {
...
+                    complete: The operation completed, all requests are processed
+                    partial: Some of the operations are successful, some failed
+                    async: Operation running asynchronously
+                    error: Operation failed because of some system error
+                 */
+                boolean allComplete = true;
+                boolean anyPartial = false;
+                boolean anyAsync = false;
+                for (JsonNode jsonNode : sourceDocuments.values()) {
+                    JsonNode jsStatus = jsonNode.findValue(STATUS);
```
jewzaam: Why is this looking at source document status?  The source documents are the documents returned, not the lightblue response.  If there happens to be a status element it is because that entity has a status element.

jewzaam: I could be mistaken, going to compare to how we do things internally to see if we're consistent.

luan-cestari: Isnt this status from the response rather the entity? I thought it is from the response for the request as described in http://jewzaam.gitbooks.io/lightblue-specifications/content/language_specification/data.html . 

Let me know if I am wrong in just the  source document status case or in the destination case too please.

luan-cestari: OK. Let me know if it is right or wrong. Also I put some TODOs that could turn into some new issues (like features that I think it should be on lightblue-client for example)

jewzaam:
I've started a new branch in my fork that attempts to resolve merge issues found in this PR: https://github.com/jewzaam/lightblue-migrator/tree/fixing-97

Additional problems around processing responses for fetching source and destination documents.  The implementations of `getSourceDocuments()` and `getDestinationDocuments()` were not modified, which means they return the data portion of the response.  This does *not* include the status of the response.  Therefore, all the unit tests fail because of [this logic](https://github.com/luan-cestari/lightblue-migrator/blob/betterStatus%2397/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java#L312-L317).  Instead of looking at why all the existing unit tests fail, this PR changes what the unit tests expect to match the new behavior.  This is wrong.  My fork is attempting to put this back together, though it is not complete.

I recommend removing the logic that looks that the status of the LightblueResponse.  If this is a concern, add it as an issue and a separate PR.  I would like to get this change in as soon as possible.  It has been in progress longer than I expected it to be.

I have some time this morning.  I'll see what I can get done in the fixing-97 branch.

luan-cestari:
I think the test should be concerned the LightblueResponse's status as it would assurance the complete flow according what we are expecting  and it is one of the main points of the migration to have the right status on each point.

I think that some of the tests might need to be to be refactored as they were not conceived with the whole scenario details that would need to run successfully , so this could be another issue/PR

jewzaam:
Replaced by https://github.com/lightblue-platform/lightblue-migrator/pull/138

PR #123, opened 2015-03-23
Fixes #121: capture source query on jobExecution

### LINKED ISSUE #121: Capture the query used to load data from the source
It's hard to debug exactly what is going on with migrations right now given it's all timestamp based.  I have a feeling that there's some nuance with timezones that is mucking with what I'm trying to debug.  So, I propose we add in the jobExecution object a way to save the query used to pull the data from the source.

jewzaam:
#122 is merged into this PR already so it builds.  Please merge #122 first.


jewzaam:
Testing this now, will move to ready for review when done.

jewzaam:
Ready for review.  Very helpful for debugging..

PR #133, opened 2015-03-25
Updated migration job test to fail if parsing json anywhere fails (in the test)

jewzaam:
Was poking around for another issue and realized if json parsing failed it would be logged to stdout but wouldn't cause the test to fail.  This update is to make the test fail so it's not some other side effect (if lucky) that causes test to fail.


dcrissman:
@jewzaam looks like this needs a merge from master

jewzaam:
merged

Issue #119, opened 2015-03-19
DAOFacade needs separate writeUpdate and writeInsert operations


paterczm:
Inserts are different from updates in that they require id of the new object to be generated. The way the migration is designed, it's the legacy system who's responsible for generating the id, which needs to be also used for persisting the object in the new system (lightblue). To support inserts, a new DAOFacade api needs to be defined.

PR #135, opened 2015-03-25
Unit tests and ID handling for DAO facade implementation

paterczm:
A number of changes in the utils submodule. I'm sure there will be more when it gets actually used.

### DIFF
--- a/utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
+++ b/utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
@@ -1,5 +1,6 @@
 package com.redhat.lightblue.migrator.utils;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -34,14 +35,36 @@ public class DAOFacadeBase<D> {
 
     private EntityIdStore entityIdStore = null;
 
+    public EntityIdStore getEntityIdStore() {
+        return entityIdStore;
+    }
+
     public void setEntityIdStore(EntityIdStore entityIdStore) {
         this.entityIdStore = entityIdStore;
+
+        try {
+            Method method = lightblueDAO.getClass().getMethod("setEntityIdStore", EntityIdStore.class);
+            method.invoke(lightblueDAO, entityIdStore);
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException("LightblueDAO needs to have a setter method for EntityIdStore", e);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        } catch (InvocationTargetException e) {
+            throw new RuntimeException(e);
+        }
     }
 
     public DAOFacadeBase(D legacyDAO, D lightblueDAO) {
         super();
         this.legacyDAO = legacyDAO;
         this.lightblueDAO = lightblueDAO;
+    }
+
+    public DAOFacadeBase(D legacyDAO, D lightblueDAO, Class entityClass) {
+        this(legacyDAO, lightblueDAO);
+        this.entityIdStore = new EntityIdStoreImpl(entityClass);
+
+        setEntityIdStore(entityIdStore);
     }
 
     private ListeningExecutorService getExecutor() {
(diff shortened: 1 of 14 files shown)


paterczm:
This addresses #119.

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
```diff
@@ -111,8 +141,7 @@ public T call() throws Exception {
         if (LightblueMigration.shouldCheckReadConsistency() && LightblueMigration.shouldReadSourceEntity()) {
             // make sure that response from lightblue and oracle are the same
             log.debug("."+methodName+" checking returned entity's consistency");
-            Method equalsMethod = lightblueEntity.getClass().getMethod("equals", Object.class);
-            if ((Boolean) equalsMethod.invoke(lightblueEntity, legacyEntity)) {
+            if (Objects.equals(legacyEntity, lightblueEntity)) {
                 // return lightblue data if they are
```
dcrissman: Why not just call lightblueEntity.equals(legacyentity)?

paterczm: Objects.equals handles null properly: http://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#equals(java.lang.Object,%20java.lang.Object)

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
```diff
@@ -31,12 +33,40 @@
...
+
+        try {
+            Method method = lightblueDAO.getClass().getMethod("setEntityIdStore", EntityIdStore.class);
+            method.invoke(lightblueDAO, entityIdStore);
+        } catch (NoSuchMethodException e) {
+            throw new RuntimeException("LightblueDAO needs to have a setter method for EntityIdStore", e);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        } catch (InvocationTargetException e) {
+            throw new RuntimeException(e);
```
dcrissman: Could this be rolled into a single catch? Something like:
```java
catch (IllegalAccessException | InvocationTargetException e){...}
```

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
```diff
@@ -31,12 +33,40 @@
...
+    public EntityIdStore getEntityIdStore() {
+        return entityIdStore;
+    }
+
+    public void setEntityIdStore(EntityIdStore entityIdStore) {
+        this.entityIdStore = entityIdStore;
+
+        try {
+            Method method = lightblueDAO.getClass().getMethod("setEntityIdStore", EntityIdStore.class);
+            method.invoke(lightblueDAO, entityIdStore);
```
dcrissman: Why the reflection?

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
dcrissman: Just curious. I am good with that.

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
dcrissman: What about providing an interface that entities need to implement? If you are requiring the method exists here anyway, it might be cleaner.

paterczm: I thought about this, but I decided for an approach which does not require any changes in the current model, entities or dao apis. It has it's disadvantages for sure.

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
paterczm: Yes, I didn't know this syntax is available in Java 7. Will change it.

utils/src/main/java/com/redhat/lightblue/migrator/utils/DAOFacadeBase.java
dcrissman: But it does require a change, right? If they don't have this method implemented than... boom. Is there a scenario that would not have setEntityIdStore invoked?

paterczm: It does not require a change in the dao interface. Your lightblueDAO implementation needs to have that setter, but it's a net new class.

Issue #108, opened 2015-03-18
Add indexes for migrationJob entity


jewzaam:
Queries are doing full scans right now.  Recommend indexes on:
* whenAvailableDate
* jobExecutions.completedFlag

jewzaam:
Pending #97 as field "jobExecutions.completedFlag" will be replaced.

dcrissman:
The above two changes are made in different branches so that we could pull the first one in by itself if desired.

PR #137, opened 2015-03-26
Issue #108 add indexes

dcrissman:
Partially addresses #108 

The first of two PRs. This one adds indexes for existing fields. We could potentially pull this in by itself, or wait for the Bill release and do them both at the same time.


jewzaam:
Bump the version so we can deploy this metadata.

dcrissman:
@jewzaam done

PR #138, opened 2015-03-27
Fixes #97: better status for job executions

### LINKED ISSUE #97: Need better status on job executions

jewzaam:
Cherry picks commits from https://github.com/lightblue-platform/lightblue-migrator/pull/120 on top of master and then adds additional things to fix unit tests and the implementation.

Replaces https://github.com/lightblue-platform/lightblue-migrator/pull/120

Fixes https://github.com/lightblue-platform/lightblue-migrator/issues/97


consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
```diff
@@ -202,10 +203,38 @@ public void run() {
...
+                    or(
+                        not(
+
+                            withSubfield("jobExecutions", withValue("jobStatus $in ['COMPLETED_SUCCESS', 'COMPLETED_PARTIAL']"))
+                        ),
+                        and(
+                            not(
+
+                                    withSubfield("jobExecutions", withValue("jobStatus $not_in ['COMPLETED_SUCCESS', 'COMPLETED_PARTIAL']"))
+                            ),
```
dcrissman: What happens for a COMPLETED_PARTIAL case? If I am reading this correctly, the migrator won't retry, so I assume it is up to the user to figure out why and schedule a retry?

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
```diff
@@ -298,17 +300,62 @@ public void run() {
...
+                    currentRun.setOverwrittenDocumentCount(totalOverwrittenDocumentCount);
+                }
+                if (allComplete) {
+                    currentRun.setJobStatus(JobStatus.COMPLETED_SUCCESS);
+                } else if (anyPartial) {
+                    currentRun.setJobStatus(JobStatus.COMPLETED_PARTIAL);
+                } else if (anyAsync) {
+                    currentRun.setJobStatus(JobStatus.COMPLETED_IGNORED);
+                } else {
+                    currentRun.setJobStatus(JobStatus.UNKNOWN);
```
dcrissman: Instead of this conditional logic, why not just create a JobStatus local variable and set that once up above?

luan-cestari: Well I created this why so it would be very clear how the status for the migration would be influenced by the status of any of the requests (I even made int eh beginning  a more detailed way which I would count the returns and etc but it would be over engineering). So Here we can just see that if you got any partial response it would have a higher priority than if you get an Async and so on (which mean that a migration job can only have one status of time and we have to deal with different documents status  )

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
luan-cestari: Yeah I would assume that would happen. Not sure if it would be helpful, but this snippet of code was made in this discussion https://github.com/lightblue-platform/lightblue-migrator/pull/120#discussion_r26861317

consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/MigrationJob.java
dcrissman: Ok, I see now. I am thinking the run method might be getting to the point that it needs to be broken up into smaller methods, but that's for another PR.

PR #139, opened 2015-03-27
fixes #108 but also depends on #97 - adds jobStatus to the MigrationJobsToExecute index

### LINKED ISSUE #108: Add indexes for migrationJob entity

dcrissman:
Add an index for jobStatus

### DIFF
--- a/consistency-checker/src/main/resources/migrationJob.json
+++ b/consistency-checker/src/main/resources/migrationJob.json
@@ -28,6 +28,10 @@
                     {
                         "dir": "$asc",
                         "field": "whenAvailableDate"
+                    },
+                    {
+                        "dir": "$asc",
+                        "field": "jobStatus"
                     }
                 ],
                 "name": "MigrationJobsToExecute",


luan-cestari:
Good catch. I will merge it.

consistency-checker/src/main/resources/migrationJob.json
```diff
@@ -28,6 +28,10 @@
                     {
                         "dir": "$asc",
                         "field": "whenAvailableDate"
+                    },
+                    {
+                        "dir": "$asc",
+                        "field": "jobStatus"
```
jewzaam: Field should be "jobExecutions.jobStatus"

Also, for mongo "The order of the fields in a compound index is very important." [1](http://docs.mongodb.org/manual/core/index-compound/).  This isn't clear, but what it means is put fields that will exclude the most document first in the index because they are queried in order of definition within in the index.

For example, if you had a Person you wanted to index on gender and first name, you put the field for first name first because it will exclude the most records faster.

Therefore, fields probably should be in this order:
* whenAvailableDate
* jobExecutions.jobStatus
* configurationName

Though based on how we use the available date I'm not sure if that's right.  We can fiddle with explain plans if it becomes a problem and figure out which is more efficient.

dcrissman: Fixed in #140

PR #144, opened 2015-03-30
remove single quotes from expression

dcrissman:
The single quotes are actually not included in the expression.


dcrissman:
@jewzaam This will need review for Ana

PR #151, opened 2015-03-31
Fixes #150: Fix processing expired jobs

### LINKED ISSUE #150: When migrator is stopped, in process jobs are not picked up
I see the incomplete jobs getting picked up by migrator, have a new execution added, but it is closed out with status of "ABORTED_DUPLICATE".  I suspect the logic for checking if a job should be run again has a few problems.  First, it shouldn't create a new execution if it's just going to abort it.  Why is it processing it at all?  Second, if it is processing the job, the old "RUNNING" or "STARTING" status execution is not updated.  It should be terminated in some way and have an end date set.  Third, I think the logic to check if a job is expired just looks for the newest actual start date without any consideration for the status or actual end date.

jewzaam:
Migrator now marks executions as dead and picks them for processing again when necessary.

### DIFF
--- a/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
+++ b/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
@@ -2,7 +2,6 @@ package com.redhat.lightblue.migrator.consistency;
 
 import static com.redhat.lightblue.client.expression.query.ArrayQuery.withSubfield;
 import static com.redhat.lightblue.client.expression.query.NaryLogicalQuery.and;
-import static com.redhat.lightblue.client.expression.query.NaryLogicalQuery.or;
 import static com.redhat.lightblue.client.expression.query.UnaryLogicalQuery.not;
 import static com.redhat.lightblue.client.expression.query.ValueQuery.withValue;
 import static com.redhat.lightblue.client.projection.FieldProjection.includeFieldRecursively;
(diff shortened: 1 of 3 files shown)


consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java
```diff
@@ -270,6 +283,24 @@ protected static boolean isJobExecutable(MigrationJob job) {
...
+    /**
+     * Update all non-complete job executions as "dead".
+     *
+     * @param job the job to cleanup
+     */
+    protected static void markRunningJobExecutionsAsDead(MigrationJob job) throws IOException {
+        int psn = 0;
+        for (MigrationJobExecution exec : job.getJobExecutions()) {
+            if (exec.getJobStatus() != null && exec.getJobStatus().isRunning()) {
+                LOGGER.info("Marking job {} execution {} as {}", job.get_id(), psn, JobStatus.COMPLETED_DEAD);
```
dcrissman: Why not just 
```java
if(exec.getJobStatus().isRunning()){...}
```

jewzaam: You're asking why check that job status isn't null?  Guess just in case.  I see it's required in metadata so that check could be removed, but why?

dcrissman: Because I read it wrong. Approving.

PR #158, opened 2015-04-07
cleanup poms

dcrissman:


dcrissman:
Seems we intermittently are getting an exception for java 8 builds. Created #164 to investigate.

PR #161, opened 2015-04-09
Sort by whenAvailableDate ascending to pick older jobs first

jewzaam:
Noticed in a dev migration that we didn't pick older jobs consistently.  Basically, we had to stop migration several times which leaves jobs in a "RUNNING" state.  We expected them to be picked up and reprocessed after those executions marked as dead.  This didn't happen until all other jobs had been migrated.  This fixes the issue and sorts when getting migrationJob documents to process.


dcrissman:
Seems we intermittently are getting an exception for java 8 builds. Created #164 to investigate.

PR #165, opened 2015-04-10
Fixes #55:  Sonar Critical Issue: Multithreaded correctness - Inconsistent synchronization

### LINKED ISSUE #55: Sonar Critical Issue: Multithreaded correctness - Inconsistent synchronization
src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyCheckerDaemon.java

```
Inconsistent synchronization of com.redhat.lightblue.migrator.consistency.ConsistencyCheckerDaemon.isThreadAliveWatcher; locked 80% of time

Multithreaded correctness - Inconsistent synchronization

The fields of this class appear to be accessed inconsistently with respect to synchronization.  This bug report indicates that the bug pattern detector judged that
The class contains a mix of locked and unlocked accesses,
At least one locked access was performed by one of the class's own methods, and
The number of unsynchronized field accesses (reads and writes) was no more than one third of all accesses, with writes being weighed twice as high as reads
A typical bug matching this bug pattern is forgetting to synchronize one of the methods in a class that is intended to be thread-safe.
You can select the nodes labeled "Unsynchronized access" to show the code locations where the detector believed that a field was accessed without synchronization.
Note that there are various sources of inaccuracy in this detector; for example, the detector cannot statically detect all situations in which a lock is held.  Also, even when the detector is accurate in distinguishing locked vs. unlocked accesses, the code in question may still be correct.
```

luan-cestari:
https://github.com/lightblue-platform/lightblue-migrator/issues/55


luan-cestari:
Removed the unnecessary "synchronization" usage. It lead to the issue https://github.com/lightblue-platform/lightblue-migrator/issues/55 and the JavaDoc of Deamon doesn't mention any requirement about locking ( https://commons.apache.org/proper/commons-daemon/apidocs/org/apache/commons/daemon/Daemon.html ).

dcrissman:
@luan-cestari would you please merge in master?

dcrissman:
Did you intend to change consistency-checker/migrationJobs.sh and add  consistency-checker/yearlyMigrationJobs.sh ?

luan-cestari:
@dcrissman merged and removed non-related changes

PR #167, opened 2015-04-14
fix comparison

dcrissman:
We have the wrong comparison.

Comparison should match: https://github.com/lightblue-platform/lightblue-migrator/blob/master/consistency-checker/src/main/java/com/redhat/lightblue/migrator/consistency/ConsistencyChecker.java#L235


dcrissman:
This is not actually a bug, the idea is you want to get the next available job that cannot yet be processed in order to determine how long to sleep.

PR #168, opened 2015-04-15
Fixes #80 : migrationJobs.sh split output into multiple chunks #80

### LINKED ISSUE #80: migrationJobs.sh split output into multiple chunks
I just created a 40meg job file, which is too large to be inserted directly (see https://github.com/lightblue-platform/lightblue-rest/issues/108). I suggest a mechanism to break this up into multiple json files.

luan-cestari:
Description and discussion on issue  https://github.com/lightblue-platform/lightblue-migrator/issues/80


dcrissman:
I am running using the example, and nothing seems to be generated. Am I doing something wrong?
> ./yearlyMigrationJobs.sh user 2014-01-01 2014-12-31 1 30000 derek63 0 5000

Also, would it make sense for the two scripts to be one?

luan-cestari:
Hi @dcrissman , I think I know what happened. I followed the suggestion from Naveen to make a second script to call the other one and store the output on a directory on the '/tmp' , so probably you expected the output somewhere else (the current directory I guess) and the script didn't mention about that. I just updated it to echo that information. Let me know if it is good for you now.

dcrissman:
The 2014 and 2015 entries overlap _id values. It would be ideal if the script were able to go right to 8736 (or greater) for 2015

...Jobs2014.json, last entry
```json
{
"_id": "userJob_8735",
"objectType": "migrationJob",
"configurationName": "user",
"startDate": "20141230T23:00:00.000-0500",
"endDate": "20141231T00:00:00.000-0500",
"whenAvailableDate": "20141231T00:01:00.000-0500",
"expectedExecutionMilliseconds" : 30000,
"jobExecutions": [],
"creationDate": "20150427T15:13:52.000-0400",
"createdBy": "derek63",
"lastUpdateDate": "20150427T15:13:52.000-0400",
"lastUpdatedBy": "derek63"
}
```

...Jobs2015.json, first entry
{
"_id": "userJob_5000",
"objectType": "migrationJob",
"configurationName": "user",
"startDate": "20150101T00:00:00.000-0500",
"endDate": "20150101T01:00:00.000-0500",
"whenAvailableDate": "20150101T01:01:00.000-0500",
"expectedExecutionMilliseconds" : 30000,
"jobExecutions": [],
"creationDate": "20150427T15:14:21.000-0400",
"createdBy": "derek63",
"lastUpdateDate": "20150427T15:14:21.000-0400",
"lastUpdatedBy": "derek63"
}
```json

File: consistency-checker/yearlyMigrationJobs.sh [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+if [ "$endDate" \> "$END_DATE" ]; then
+    endDate=$END_DATE
+    runOnce=1
+fi
+currentYear=$(date +%Y -d $START_DATE)
+nextYear=$((currentYear+1))
+
+i=$STARTING_ID
+while true; do
+    sh migrationJobs.sh $ENTITY_NAME $startDate $endDate $NUM_HOURS $EXPECTED_EXECUTION_TIME $CREATED_BY $i > /tmp/jobs/termsAcknowledgementJobs${currentYear}.json
```
dcrissman: Output files should not be prefixed with termsAcknowledgementJobs

luan-cestari:
@dcrissman . About the overlap _id values, I can fix that but it will start to make a very complex script, which I thought it was not intended to be this was in the beginning as the it was suggest to jump to the next range of ids (in https://github.com/lightblue-platform/lightblue-migrator/issues/80#issuecomment-83615208  it would skip 10000 ids for each iteration, so the user should be aware if it would overlap and use a bigger number if necessary). So, if you this feature, to make it consistent I would suggest to convert the script into a java project , so we could have a set of unit test for those cases and other features. Do you agree with me to make the project?

dcrissman:
I don't know that I would be in favour of making it a new top level project. Switching to Java I am not opposed too, if it makes sense.

Couldn't the script just keep track of the last id between files?

luan-cestari:
@dcrissman done

migrator/yearlyMigrationJobs.sh
```diff
@@ -0,0 +1,65 @@
...
+if [ "$endDate" \> "$END_DATE" ]; then
+  endDate=$END_DATE
+  runOnce=1
+fi
+currentYear=$(date +%Y -d $START_DATE)
+nextYear=$((currentYear+1))
+
+i=$STARTING_ID
+while true; do
+  sh migrationJobs.sh $ENTITY_NAME $startDate $endDate $NUM_HOURS $EXPECTED_EXECUTION_TIME $CREATED_BY $i > /tmp/jobs/termsAcknowledgementJobs${currentYear}.json
```
dcrissman: Please replace "termsAcknowledgementJobs" with $ENTITY_NAME

dcrissman:
The _id values still seem to collide. What about adding the year to the _id, and then you could simply restart the count with each year?

ex. userJob_1_2015

PR #182, opened 2015-05-05
Rename consistency-checker to migrator

bserdar:


dcrissman:
There were some concerns about this regarding complications to a future releases. Does anyone have any specific concerns or thoughts they would like to add?

There are essentially two "projects' in this repo, the migrator itself and the utility library applications can use to progressively switch traffic over to lightblue away from their legacy backends. I just want to make sure these two "projects" are each adequately represented.

bserdar:
In that case, lets make it explicit:
 - merge core and utils. We can name that package "migration-utils"
 - The new migrator subtree remains. I don't see a problem with consistency
checker living in the same package as the migrator
 - For JDBC migrations: instead of writing those into this project, lets
have a separate project where we develop JDBC migrations outside the
migrator package.  We can write some sort of a glue class that makes the
migrator aware of the JDBC migrators.
 - Later, we can get rid of JDBC migrators, but even then, we can keep the
actual migration logic outside the migrator package

dcrissman:
@bserdar would you please merge in changes from master?

bserdar:
done.

PR #191, opened 2015-05-14
New main

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/Main.java
```diff
@@ -1,81 +1,111 @@
...
+    public static ConsistencyChecker buildConsistencyChecker() {
+        ConsistencyChecker checker = new ConsistencyChecker();
+        checker.setConsistencyCheckerName(System.getProperty("name"));
+        checker.setHostName(System.getProperty("hostname"));
+        checker.setConfigPath(System.getProperty("config"));
+        checker.setMigrationConfigurationEntityVersion(System.getProperty("configversion"));
+        checker.setMigrationJobEntityVersion(System.getProperty("jobversion"));
+        checker.setSourceConfigPath(System.getProperty("sourceconfig"));
+        checker.setDestinationConfigPath(System.getProperty("destinationconfig"));
+        return checker;
```
dcrissman: Why system properties over a configuration file?

migrator/src/main/resources/log4j.properties
```diff
@@ -11,7 +11,7 @@ log4j.rootLogger=INFO, console
 
 # LOG FILE LOCATIONS
 log4j.appender.console=org.apache.log4j.ConsoleAppender
-log4j.appender.console.Threshold=INFO
+log4j.appender.console.Threshold=DEBUG
 log4j.appender.console.layout=org.apache.log4j.PatternLayout
```
dcrissman: Do we want to change the log level?

migrator/src/main/java/com/redhat/lightblue/migrator/Main.java
bserdar: That is your code. I only moved it to a separate function.

migrator/src/main/resources/log4j.properties
bserdar: No, that will be reverted back. I set that to debug to see the test logs.

migrator/src/main/java/com/redhat/lightblue/migrator/Main.java
dcrissman: Oh, I see it. I don't think I originated it, but I do see where you just kept the existing code. Puppet is passing in those parameters as jvm arguments I think. You can disregard this comment.

Issue #40, opened 2014-12-05
Add configurability to how many times the daemon would attempt to restart the Migrator thread.


dcrissman:
This is a spin off of #38 38. Some of the logic has already been roughed in, but in the spirit of getting a merge accepted, we decided to break out into its own feature.

The desire is that a optional configuration option can be passed into the daemon stating how many time it would attempt to restart the worker thread before exiting out altogether. It should support -1 for infinite retries (default), 0 for don't retry at all, or an int specifying the max number of retries. All events should be logged.

jewzaam:
Could be a hard throttle or maybe a delayed restart after some number of failures?

bserdar:
No longer relevant

PR #193, opened 2015-05-18
Enhancement for thread management (specially the shutdown)

luan-cestari:
I was working on  issue https://github.com/lightblue-platform/lightblue-migrator/issues/40 , but after the commit https://github.com/lightblue-platform/lightblue-migrator/commit/23e7ccd6c377669fd39927ad5bc7be4f17733db8 the migrator doesn't have a retry mechanism anymore. I noticed that I lack a graceful shutdown of the thread pool created. I still have some concerns of the thread pools created , which could be just a single pool or that the pool used seems to be wasting resource after running all jobs and other possible unwanted scenarios.


bserdar:
Please do not work on migrator core. There are serious threading problems
and race conditions that i am working to fix.

PR #195, opened 2015-05-22
Facade measures and logs source and dest method execution times

paterczm:
Gives more insight into response times.


paterczm:
Does not build because of a problem in migrator (related to recent changes in lightblue-client).

bserdar:
Dont' worry about it, i'll merge in when i'm done with migrator changes.

PR #197, opened 2015-05-28
Add controller

bserdar:


dcrissman:
This PR is not yet ready to be merged, but has been pushed out here so that we can start reviewing the changes.

bserdar:
If you want to review this, it is much easier if you start reading from Main in my local branch, and go deeper from there. Although the original load/compare/save bits are retained in some form, they are not where they used to be, and a lot of other things are written from scratch, making diff-browsing incomprehensible.

migrator/src/main/java/com/redhat/lightblue/migrator/Breakpoint.java
```diff
@@ -0,0 +1,116 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Works like a breakpoint in a debugger. If stop is called, execution
+ * stops when run() is called until resume() is called by another
+ * thread.
+ */
+public class Breakpoint {
```
dcrissman: Is the intention of this class for debugging purposes?

bserdar: It is used in unit tests, to stop the execution of a thread at a certain
point, inspect thread state, and then resume exection.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
```diff
@@ -0,0 +1,121 @@
...
+    
+    public LightblueClient getLightblueClient() {
+        LightblueHttpClient httpClient;
+        if (cfg.getClientConfig() != null) {
+            httpClient = new LightblueHttpClient(cfg.getClientConfig());
+        } else {
+            httpClient = new LightblueHttpClient();
+        }
+        return new LightblueHystrixClient(httpClient, "migrator", "primaryClient");
+    }
```
dcrissman: Any reason not to lazy load and just return the class instance if not null? The method name getLightblueClient() makes me think I would get the same client instance each time.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    @Override
+    public void run() {
+        LOGGER.debug("Starting controller");
+        boolean interrupted=false;
+        Breakpoint.checkpoint("Controller:start");
+        while(!interrupted) {
+            interrupted=Thread.interrupted();
```
dcrissman: Sanity check. Thread.interrupted will clear the interrupted state, while isInterrupted will not clear the state. I guess my question is, do we want to clear the state here or not?

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
```diff
@@ -0,0 +1,121 @@
...
+                    createControllers(cfg);
+                    Breakpoint.checkpoint("Controller:createconfig");
+                } catch (Exception e) {
+                    LOGGER.error("Error during configuration load:"+e);
+                }
+            }
+            if(!interrupted) {
+                try {
+                    Thread.sleep(30000);
+                } catch (Exception e) {}
```
dcrissman: My understanding is once a InterruptedException is caught, the status is cleared. So the interrupted=Thread.interrupted(); above would return false.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
dcrissman: Also, should this check be at the end of the loop? Otherwise an interrupted thread would go through one more iteration.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
bserdar: This is called from the child threads, so each thread gets a new instance
of the client. This is done so we don't make any assumptions about the
thread safety of the LightblueClient.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
bserdar: That's why result of interrupted is used to set the flag value, but callign
isInterrupted should be good as well. I'll change it.

bserdar: all code blocks are enclosed with if(!interrupted), so once interrupted, it
should fall through without doing anything.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
bserdar: Catching InterruptedException won't clear interrupted state.

dcrissman: http://stackoverflow.com/questions/2523721/why-do-interruptedexceptions-clear-a-threads-interrupted-status

dcrissman:
You are missing the 'o' in Job, but I assume you plan to remove anyway? migrator/src/main/java/com/redhat/lightblue/migrator/MigrationJb.java.old

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
bserdar: Should be fixed in a1152df2d65165373af461d938aecbe9c4917038

bserdar:
Yes, that'll be removed eventually.

migrator/src/main/resources/migrationJob.json
```diff
@@ -254,4 +209,4 @@
             "value": "1.0.0"
```
dcrissman: Should probably be a version bump.

bserdar: Ok.

migrator/src/main/java/com/redhat/lightblue/migrator/Breakpoint.java
dcrissman: Interesting, ok.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
dcrissman: We could call it createLightblueClient, but I wouldn't block because of that.

bserdar: Makes sense, i'll change

migrator/src/main/resources/migrationJob.json
```diff
@@ -60,16 +41,16 @@
...
-                "lb-migrationjob-find"
+                "anyone"
             ],
             "insert": [
-                "lb-migrationjob-insert"
+                "anyone"
             ],
             "update": [
-                "lb-migrationjob-update"
+                "anyone"
```
dcrissman: Do we want anyone to be able to make changes to migrationJob?

migrator/src/main/resources/migrationConfiguration.json
```diff
@@ -47,16 +33,16 @@
...
-                "lb-migrationconfig-find"
+                "anyone"
             ],
             "insert": [
-                "lb-migrationconfig-insert"
+                "anyone"
             ],
             "update": [
-                "lb-migrationconfig-update"
+                "anyone"
```
dcrissman: Do we want anyone to be able to make changes to migrationConfiguration?

migrator/src/main/resources/migrationConfiguration.json
```diff
@@ -2,24 +2,10 @@
...
-                    "delete"
-                ],
-                "configuration": {
-                    "entityName": "audit",
-                    "version": "1.0.0"
-                },
-                "name": "auditHook"
-            }
-        ],
+        "defaultVersion" : "1.0.0",
```
dcrissman: Should probably bump this version also.

migrator/src/main/resources/log4j.properties
```diff
@@ -7,11 +7,11 @@
 # LOGGING LEVELS
 
 # To turn more verbose logging on - change "WARN" to "DEBUG"
-log4j.rootLogger=INFO, console
+log4j.rootLogger=DEBUG, console
 
 # LOG FILE LOCATIONS
 log4j.appender.console=org.apache.log4j.ConsoleAppender
-log4j.appender.console.Threshold=INFO
+log4j.appender.console.Threshold=DEBUG
```
dcrissman: debug logging?

migrator/src/main/resources/activeExecution.json
```diff
@@ -0,0 +1,107 @@
...
+                "anyone"
+            ],
+            "find": [
+                "anyone"
+            ],
+            "insert": [
+                "anyone"
+            ],
+            "update": [
+                "anyone"
```
dcrissman: Do we want anyone to be able to make changes to activeExecution?

migrator/src/main/resources/migrationJob.json
bserdar: Did that so I can test. Will revert later.

migrator/src/main/resources/log4j.properties
bserdar: Will change, i'm stll testing this thing.

PR #213, opened 2015-06-18
Fix #184: Adding timeout to lightblue calls

### LINKED ISSUE #184: DAOFacade should support timeouts for calls to destination service (lightblue)
This change will ensure that clients aren't affected too much by potential performance problems in Lightblue. If Lightblue fails to respond in a predefined time frame, DAOFacade should drop the connection, log a warning and return results from the legacy service.

paterczm:
Fixes https://github.com/lightblue-platform/lightblue-migrator/issues/184.

Also some refactoring and logging improvements.


bserdar:
What happens when a call times out?

paterczm:
Facade returns data from legacy dao (oracle). The call to lightblue continues in a separate, detached thread.

bserdar:
If things slow down, update requests may start piling up. What do we do
then?

PR #232, opened 2015-07-20
Fix comparison issues in BeanConsistencyChecker

ykoer:
Handle BigDecimal comparison by ignoring the decimal scale. 
Use Object.deepEquals instead Object.equals since it didn't work with object arrays


paterczm:
Looking good, but please add a test with an array.

I'm also seeing migrator tests failing. It could be related to the recent changes in lightblue client.

bean-consistency-checker/src/test/java/com/redhat/lightblue/migrator/consistency/BeanConsistencyCheckerTest.java
```diff
@@ -64,6 +66,29 @@ public void testInconsistencyList() {
...
+        Country[] cArr1 = new Country[] { new Country(1l, "PL"), new Country(2l, "CA")};
+        Country[] cArr2 = new Country[] { new Country(3l, "PL"), new Country(2l, "CA")};
+
+        Assert.assertFalse(Objects.deepEquals(cArr1, cArr2));
+
+        Country[] cArr3 = new Country[] { new Country(2l, "CA"), new Country(1l, "PL")};
+        Country[] cArr4 = new Country[] { new Country(1l, "PL"), new Country(2l, "CA")};
+
+        Assert.assertFalse(Objects.deepEquals(cArr3, cArr4));
+    }
```
paterczm: The point is to test beanConsistencyChecker.consistent method, not Objects.deepEquals.

dcrissman:
@ykoer @paterczm - I just re-ran the build for the master branch and it did indeed fail. It is producing a lot of output, the quantity of logs is actually what made travis fail. Locally, I get a LOT of output also, but the build passes. Something isn't working.

paterczm:
@dcrissman: When I build it locally, it's logging the same exceptions, but it's nowhere near 4 MB (what travis reports). Maybe it's getting stuck.

PR #236, opened 2015-08-12
Facade handles dao exceptions properly

paterczm:
* exceptions are not wrapped in java.lang.reflect.InvocationTargetException anymore
* Throwable is used instead of Exception so Errors can be handled as well
* updated example CountryDAO to use CountryException


paterczm:
@bserdar, this is to address the problem you identified in qa

PR #241, opened 2015-08-19
Add a new flag to migration job to mark auto generated jobs

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/MigrationJob.java
```diff
@@ -28,6 +28,8 @@
 
     private String status; // available, processing, complete
 
+    private boolean generated; // Set if generated by consistency checker
```
dcrissman: Why not do this as javadoc? (I won't block on this)

PR #243, opened 2015-08-24
Json compare

bserdar:
Replaces the entity comparison algorithm with skyscreamer.JSONCompare


migrator/src/main/java/com/redhat/lightblue/migrator/Utils.java
```diff
@@ -68,68 +79,109 @@ public static LightblueClient getLightblueClient(String configPath)
...
-                compareDocs(inconsistentPaths, entry.getValue(), destinationDocument.get(entry.getKey()),
-                            StringUtils.isEmpty(path) ? entry.getKey() : path + "." + entry.getKey(),excludes);
+    private static boolean matches(String path,String pattern) {
+        String[] patternComponents=pattern.split("\\.");
+        String[] pathComponents=path.split("\\.");
+        if(patternComponents.length<=pathComponents.length) {
+            for(int i=0;i<patternComponents.length;i++) {
+                if(pathComponents[i].equals(patternComponents[i])||
+                   "*".equals(patternComponents[i]))
+                    ;
```
dcrissman: Looks like this if statement doesn't do anything. Should the conditional be reversed?

dcrissman:
I would really like to see a description summarizing what this change is.

dcrissman:
Totally as an aside, I noticed you like to use one line conditions or loops without curly braces. You do indent properly, so I'll give you that, but in general I am not a fan of this practice. I think the potential for it to be misinterpreted is not worth the handful of characters that it saves. But that is just my 2 cents.

migrator/src/main/java/com/redhat/lightblue/migrator/Inconsistency.java
```diff
@@ -0,0 +1,61 @@
+package com.redhat.lightblue.migrator;
+
+import java.util.List;
+
+public class Inconsistency {
```
dcrissman: JavaDoc?

bserdar: Added

bserdar:
Added.

migrator/src/main/java/com/redhat/lightblue/migrator/Utils.java
bserdar: Reversed.

dcrissman:
@bserdar looks like the build failed. Fix and I will merge.

PR #244, opened 2015-08-25
Check if migration jobs exists, and don't run consistency checker if …

bserdar:
…they do

### DIFF
--- a/migrator/src/main/java/com/redhat/lightblue/migrator/ConsistencyCheckerController.java
+++ b/migrator/src/main/java/com/redhat/lightblue/migrator/ConsistencyCheckerController.java
@@ -152,7 +152,7 @@ public class ConsistencyCheckerController extends AbstractController {
                     period=parsePeriod(migrationConfiguration.getPeriod());
 
                     if(migrationJobsExist()) {
-                        LOGGER.error("There are migration jobs for {}, not running consistency checker this time",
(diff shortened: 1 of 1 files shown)


migrator/src/main/java/com/redhat/lightblue/migrator/ConsistencyCheckerController.java
```diff
@@ -131,67 +151,72 @@ public void run() {
...
-                    try {
-                        ae=lock(lockId);
-                    } catch (Exception e) {
-                        LOGGER.error("Exception during lock attempt {}:{}",lockId,e);
-                        ae=null;
-                    }
-                    if(ae!=null) {
-                        Breakpoint.checkpoint("CCC:locked");
+                    if(migrationJobsExist()) {
+                        LOGGER.error("There are migration jobs for {}, not running consistency checker this time",
```
paterczm: Is this really an error?

bserdar: Should I make it an info?

paterczm: This is normal operation, right? Info should be fine.

bserdar: It is info now.

PR #245, opened 2015-08-27
Handling inconsistency logging for methods with primitive arrays

paterczm:
This is a fix to ```java.lang.ClassCastException: [J cannot be cast to [Ljava.lang.Object;``` exceptions when logging inconsistencies for methods with arrays of primitives as parameters.

Having to handle primitives and objects differently sucks. One more reason to use scala.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
@@ -1,5 +1,6 @@
 package com.redhat.lightblue.migrator.facade;
 
+import java.lang.reflect.Array;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
```diff
@@ -138,22 +139,39 @@ private ListeningExecutorService createExecutor() {
...
-        Iterator<Object> it = Arrays.asList(values).iterator();
-        while(it.hasNext()) {
-            Object value = it.next();
-            if (value != null && value.getClass().isArray())
-                str.append(Arrays.deepToString((Object[])value));
-            else
-                str.append(value);
-            if (it.hasNext()) {
-                str.append(", ");
+    static class MethodCallStringifier {
```
bserdar: Why in a class?

paterczm: To be able to unit test without initializing DAOFacadeBase.

paterczm: Though I guess a static method would do too.

PR #248, opened 2015-09-04
Use locking, some cleanup

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/AbstractController.java
```diff
@@ -17,6 +18,8 @@
 public abstract class AbstractController extends Thread {
 
     private static final Logger LOGGER=LoggerFactory.getLogger(AbstractController.class);
+
+    private final String callerId=UUID.randomUUID().toString();
```
dcrissman: Locking will use a random UUID by default. No reason to duplicate that effort.

bserdar: Forgot that it does. Will remove.

migrator/src/test/resources/simplelogger.properties
```diff
@@ -1,3 +1,3 @@
 org.slf4j.simpleLogger.defaultLogLevel=ERROR
-org.slf4j.simpleLogger.log.com.redhat.lightblue.migrator=ERROR
+org.slf4j.simpleLogger.log.com.redhat.lightblue.migrator=DEBUG
```
dcrissman: I assume this was for testing purposes, but we probably don't actually want it this way.

bserdar: These are fixed now.

PR #250, opened 2015-09-10
Fix #249: Handle exceptions properly in lightblue proxy phase of the migration 

### LINKED ISSUE #249: During proxy phase, facade tries to return legacy data on lightblue error or timeout
Facade is configured to fall back to legacy on lightblue failure on timeout. This does not make sense when migration is over and lightblue is the only system available. At that point facade should simply throw lightblue exceptions if they occur.

paterczm:
There are no unit tests for timeout scenario, but timeouts don't make sense in the proxy phase (when only lightblue is available), so I just assume timeouts will be turned off.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
@@ -208,6 +208,15 @@ public class DAOFacadeBase<D> {
         }
     }
 
+    private Throwable extractUnderlyingException(Throwable t) {
+        if ((t instanceof ExecutionException || t instanceof InvocationTargetException) && t.getCause() != null) {
+            return extractUnderlyingException(t.getCause());
+        }
+        else {
+            return t;
+        }
+    }
+
     /**
      * Call dao method which reads data.
      *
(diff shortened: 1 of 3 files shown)


paterczm:
@ykoer, this is what we discussed today.

facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
```diff
@@ -208,6 +208,15 @@ public T call() throws Exception {
         }
     }
 
+    private Throwable extractUnderlyingException(Throwable t) {
```
ykoer: Apache Commons Lang is already in the classpath, so you could use ExceptionUtils.getRootCause(throwable) instead this. I think it also traverses the exception chain to find the last one.

paterczm: I don't want to get to the root cause, I just want to find the exception returned by the service. It may be a chain of causes, just stripped from exception wrappers added by the facade (ExecutionException and InvocationTargetException).

paterczm: It's not perfect, because if the service returns ExecutionException or InvocationTargetException, facade will tamper with it. Do you think I should make it more strict?

PR #252, opened 2015-09-23
Fixing lightblue-migrator project cleanup

paterczm:
Restoring some missing classes, deleting web module for good.


paterczm:
Basically, this is to fix ee8d86b748f4083b8c86dde305d4bd70a9549f11.

facade/src/main/java/com/redhat/lightblue/migrator/features/TogglzRandomUserProvider.java
```diff
@@ -0,0 +1,21 @@
...
+import org.togglz.core.user.UserProvider;
+
+/**
+ * Call {@link TogglzRandomUsername#init()} to generate new random username before
+ * this provider is used. {@link DAOFacadeBase} apis already handle that.
+ *
+ * @author mpatercz
+ *
+ */
+public class TogglzRandomUserProvider implements UserProvider {
```
dcrissman: Is this for testing purposes?

paterczm: No. It's for creating usernames which remain the same for the duration of the api call. I'm abusing Togglz here, it's not very suitable for request based gradual rollout.

PR #260, opened 2015-10-28
Fix #259: Add implementation name to inconsistency log

### LINKED ISSUE #259: There is no easy way to tell which service is generating the inconsistencies
Facade is logging exact method where inconsistencies were found, but it does not log the service name. It's impossible to generate reports on number of inconsistencies based on a report.

paterczm:
This will make creating reports on inconsistencies per service much easier.

https://github.com/lightblue-platform/lightblue-migrator/issues/259


paterczm:
Is this the last catch?
```java
        } catch (JsonProcessingException e) {
            log.error("Consistency check failed! Invalid JSON. " + e.getMessage());
        }
```

PR #261, opened 2015-11-02
Don't try to relock what we already locked

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/AbstractController.java
```diff
@@ -89,6 +93,7 @@ public void unlock(String id) {
         } catch (Exception e) {
             LOGGER.error("Error unlocking {}",id,e);
         }
+	myLocks.remove(id);
```
paterczm: Is it possible to leak those locks? Maybe use finally when removing a lock from myLocks?

bserdar: The possible exceptions are already caught.

PR #263, opened 2015-11-03
Add custom json comparator

bserdar:


paterczm:
There are unit tests only for following combination:
```java
diff.setOption(JsonDiff.Option.ARRAY_ORDER_INSIGNIFICANT);
diff.setOption(JsonDiff.Option.RETURN_LEAVES_ONLY);
```
Is it because we don't need other modes of diff? If so, maybe we don't need those options at all?

bserdar:
There are tests for ARRAY_ORDER_SIGNIFICANT.

There aren't for RETURN_PARENT_DIFFS. I'll add some.

Issue #171, opened 2015-04-16
Having a Proxy Facade Factory would make migration & testing efforts easier


jblashka:
Currently during migration we have to write a DAOFacadeImpl for each and every DAO, but the contents of each method is basically the same:
return callDAOXMethod(ReturnType[].class, "MethodName", new Class[]{ArgumentType.class}, Argument);

Instead, we can create a proxy factory that takes in the two implementations and takes care of all the grunt work of writing all these methods. We can add annotations (@Read, @Write, @SourceOnly, etc.) to the DAO interface that the proxy's interceptor can utilize when methods are called to determine how to handle the method calls. This would be a new class utilizing the existing DAOFacadeBase in case we still need to have some exceptions in how the method calls are handled we'll still have the old process to fall back on.

This also lightens the testing effort because there's no longer many FacadeDAOImpl classes to test, just the proxy factory.

paterczm:
+1

The annotations we need are: Read,Insert, Update and SourceOnly.

dcrissman:
I would argue that we still want to test the api endpoints against a lightblue backend and traditional backend to ensure the same results.

PR #269, opened 2015-11-10
Init facade as dynamic proxy (using InvocationHandler)

paterczm:
This is https://github.com/lightblue-platform/lightblue-migrator/issues/171 implementation. The benefit is less code to write and maintain. Example:

Current solution:
```java
public class DAOFacadeExample extends DAOFacadeBase<CountryDAO> implements CountryDAO {		
			
    public DAOFacadeExample(CountryDAO legacyDAO, CountryDAO lightblueDAO) {		
        super(legacyDAO, lightblueDAO);		
    }		
		
    @Override		
    public Country createCountry(Country country) throws CountryException {		
        try {		
            return callDAOCreateSingleMethod(entityIdExtractor, Country.class, "createCountry", country);		
        } catch (CountryException ce) {		
            throw ce;		
        } catch (Throwable t) {		
            throw new RuntimeException(t);		
        }		
    }

   ( ... )

```

Dynamic proxy solution:
```java
public class DAOFacadeExample implements CountryDAO {

    CountryDAO facade;

    public DAOFacadeExample(CountryDAO legacyDAO, CountryDAO lightblueDAO) {		
        facade = FacadeProxyFactory.createFacadeProxy(legacyDAO, lightblueDAO, );
    }		
		
    @Override		
    public Country createCountry(Country country) throws CountryException {		
        return facade.createCountry(country);
    }

   ( ... )
```

The facade knows which operation this is based on the ReadOperation, WriteOperation and UpdateOperation annotations which need to be added to CountryDAO apis. No annotation means the call will be sent to legacyDAO, no matter the togglz configuration.

This change is backwards compatible, you can still extend DAOFacadeBase if you prefer.


paterczm:
@alechenninger, @jblashka, I think you guys proposed this originally. Please take a look.

facade/src/main/java/com/redhat/lightblue/migrator/facade/proxy/FacadeProxyFactory.java
```diff
@@ -0,0 +1,121 @@
...
+     * {@link DAOFacadeBase#callDAOCreateSingleMethod(EntityIdExtractor, Class, String, Class[], Object...)} for details.
+     *
+     * @author mpatercz
+     *
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    public static @interface WriteSingleOperation {
+        String entityIdExtractorClassName();
+    }
```
bserdar: How about Class, instead of String className?

PR #270, opened 2015-11-11
Don't sort migrator job resultset, it exceeds mongo internal limits w…

bserdar:
…hen there are too many jobs


ykoer:
It means we don't care anymore about the job order?

bserdar:
Not really, as long as its scheduled date is expired, it runs.

All our jobs have the same expiration date.

PR #276, opened 2015-11-12
Service Facade

paterczm:
This is a rewrite of lightblue-facade to support passing data between services and singleton operations (like sending email). The new facade is not backwards compatible. List of changes:
* SharedStore to pass data between the legacy and lightblue service. Any kind of data. What exactly is being passed and retrieved is part of the service implementation, the facade just provides the SharedStore (the services need to implement SharedStoreSetter interface). Passing data makes sense only for serial calls (facade calls legacy service first, then lightblue service). Parallel calls cannot pass data. By default, all calls are serial, but this can be controlled in the annotation, e.g. ReadOperation(parallel=true).
* Renamed classes and methods not to suggest this is a DAO only facade
* Unified the read, write and update calls, that's less code to maintain


paterczm:
Added back old facade for backwards compatibility, but DAOFacadeBase is duplicated code. I could refactor it to use the new code to some extend, but it's not possible to do for the write method. Id handling logic changed too much.

PR #277, opened 2015-11-16
Facade should not log huge inconsistencies into server.log

ykoer:
Because server.log is sent to Splunk.

Logic proposal:
1. Build inconsistency log: diff, source and destination
2. If it's < 64kB, log it into server.log (no change in behavior)
3. If it's > 64kB, build new inconsistency log: diff only
4. If it's < 64kB, log it into server.log
5. If it's still > 64kB, log in server.log only where the inconsistency was found and log full inconsistency in a separate logger at debug level. We can then adjust logging configuration at the server level to log it into a separate file.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
@@ -7,11 +7,15 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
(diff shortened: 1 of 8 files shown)


paterczm:
DAOFacadeBase is deprecated now, but some services will continue using it. Please add your changes to ServiceFacade (new implementation) too.

facade/src/main/java/com/redhat/lightblue/migrator/facade/DAOFacadeBase.java
```diff
@@ -49,6 +49,7 @@
 public class DAOFacadeBase<D> {
 
     private static final Logger log = LoggerFactory.getLogger(DAOFacadeBase.class);
+    private static final Logger logInconsisteny = LoggerFactory.getLogger("Inconsistency");
```
paterczm: Typo in const name.

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -48,6 +48,7 @@
 public class ServiceFacade<D> implements SharedStoreSetter {
 
     private static final Logger log = LoggerFactory.getLogger(ServiceFacade.class);
+    private static final Logger logInconsisteny = LoggerFactory.getLogger("Inconsistency");
```
paterczm: Same typo.

paterczm:
Perhaps you could extract the inconsistency logging logic to avoid redundancy in both ServiceFacade and DAOFacadeBase? It will pay off if there are bugs to fix.

ykoer:
OK

paterczm:
@ykoer, one more important thing. The logs may contain sensitive information (like personal information). I suggest to add a switch to the facade to log data and diffs to separate log. You already implemented that logic depending on size, so should be very easy to add.

PR #295, opened 2016-01-19
Do not timeout destination calls in lightblue proxy phase

paterczm:
Facade timeout isolates users from potential lightblue performance problems. Once source system is disabled, it doesn't make sense to timeout lightblue calls anymore.


paterczm:
Note to reviewer: this bug fix is implemented twice - for ServiceFacade and the legacy DAOFacadeBase.

PR #296, opened 2016-01-22
Various updates

jewzaam:
* updated versions of dependencies
* fixed resource not getting closed
* fixed some race / flow control issues in Breakpoint
* fixed one API that had params swapped

### DIFF
--- a/migrator/src/main/java/com/redhat/lightblue/migrator/Breakpoint.java
+++ b/migrator/src/main/java/com/redhat/lightblue/migrator/Breakpoint.java
@@ -51,8 +51,10 @@ public class Breakpoint {
     public void resume() {
         if(stopped) {
             synchronized(this) {
-                stopped=false;
-                notify();
+                if (stopped) {
+                    stopped=false;
+                    notify();
+                }
             }
         }
     }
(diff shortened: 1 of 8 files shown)


migrator/src/main/java/com/redhat/lightblue/migrator/Breakpoint.java
```diff
@@ -86,8 +88,10 @@ public void checkpoint() {
         } else {
             if(waiting) {
                 synchronized(this) {
-                    ran=true;
-                    notify();
+                    if (waiting) {
```
bserdar: waiting needs to be volatile for this to work, could you change?

jewzaam: So if waiting is true before synchronized but false after, it should still notify?

bserdar: 'waiting' isn't declared as 'volatile'. It should be volatile, it doesn't change whether notify is called or not.

jewzaam: updated

PR #301, opened 2016-02-04
Generating method call string only when needed (i.e. when there is an…

paterczm:
… inconsistency to log). It was generated for every call before, that's unnecessary.


facade/src/main/java/com/redhat/lightblue/migrator/facade/methodcallstringifier/MethodCallStringifier.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Generates method(arg1, arg2, ...) string used to trace errors and data inconsistency warnings to particular
+ * method call.
+ *
+ * @author mpatercz
+ *
+ */
+public interface MethodCallStringifier {
+
```
bserdar: Ugly name, technically you don't need it, you can simply use an Object

PR #313, opened 2016-03-08
Consistency check performance test

paterczm:
@bserdar 

Jiff is ~4 times faster than ConsistencyChecker implementation using JSONCompare. As far as I remember, we're not using Jiff in the facade because it does not do pretty diffs. How much effort would it be to change that?

We could also use both methods:

- Jiff to check for consistency and then JSONCompare in a new thread only if there is inconsistency 
- No need to worry about ignoring fields, since this is handled during serialization to json. Both Jiff and JSONCompare would use the same settings to ignore fields, there is no risk of those being different.

Thoughts?

### DIFF
--- a/facade/pom.xml
+++ b/facade/pom.xml
@@ -85,6 +85,12 @@
         <artifactId>togglz-junit</artifactId>
         <scope>test</scope>
     </dependency>
+    <dependency>
+        <groupId>com.redhat.lightblue.migrator</groupId>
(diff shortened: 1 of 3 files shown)


bserdar:
There is another comparator in core that prints out nice diffs. There is an
algorithm in lightblue-migrator that compares using Jiff, and if it is
different, compares again using comparator to print a nice diff. We can use
the same thing.

Issue #308, opened 2016-02-24
ServiceFacade requires implementation of SharedStoreSetter


kahowell:
Right now even if the implementations don't use the `SharedStore`, they need to implement it. I think this could be avoided if we simply checked if the implementations implemented `SharedStoreSetter` (via `instanceof` in `ServiceFacade.setSharedStore`), then it wouldn't require this.

Alternatively, if we do want to force this interface on implementations, we could make this a compile-time error by having the `SharedStoreSetter`s be specified separately from the services in the constructor (ex.  `ServiceFacade(D legacySvc, SharedStoreSetter legacySharedStoreImpl, D lightblueSvc, SharedStoreSetter lightblueSharedStoreImpl, Class serviceClass)`.

paterczm:
If you do use SharedStore, but forget to implement SharedStoreSetter, then things will not work in a non obvious way. We run into this and that's why it's required. I agree it would be better if it failed at compilation time, rather than deployment. I could come up with bounded types based solution, but it would be a client breaking change. I'll think about it.

PR #314, opened 2016-03-10
Require SharedStoreSetter inferface on facaded services at compile time

paterczm:
https://github.com/lightblue-platform/lightblue-migrator/issues/308

This will break the services. Don't merge until we know we want this.


kahowell:
Cool! :+1: Thanks for this.

PR #315, opened 2016-03-10
ConsistencyChecker performance improvements

paterczm:
Large result sets run in MBs. Performing consistency checks can take really long (hundreds ms). Also, scanning classpath for ModelMixIn classes adds 70 ms of additional overhead (at least on my laptop).

Changes:

- Use Jiff for initial consistency check. It's faster than JSONCompare, but does not handle diffs as well (?). If objects are inconsistent, use JSONCompare to generate and log the diff in a non blocking manner.
- Create ObjectMapper for pojo2json conversion on ConsistencyChecker initialization. To be able to do that, removed includeMethods functionality, which allows to limit field ignore to certain methods only. It was introduced for performance reasons, but hopefully those reasons are gone now, after changes explained in the previous point. **Removing includeMethods is a breaking change**.

### DIFF
--- a/facade/pom.xml
+++ b/facade/pom.xml
@@ -111,6 +111,9 @@
           <excludes>
             <exclude>**/ConsistencyCheckPerforformanceTest.java</exclude>
           </excludes>
+          <systemPropertyVariables>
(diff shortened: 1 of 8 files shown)


paterczm:
@ykoer, please also take a look at this.

PR #331, opened 2016-04-11
add isEmpty check

dcrissman:
Add the ability to check if the something is currently in the shared store.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/sharedstore/SharedStore.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/sharedstore/SharedStore.java
@@ -44,4 +44,9 @@ public interface SharedStore {
      */
     public void clear();
 
+    /**
+     * @return true if a value is currently cached, otherwise false.
+     */
+    public boolean isEmpty();
+
 }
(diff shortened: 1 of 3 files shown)


paterczm:
@dcrissman, what is it for?

dcrissman:
I have a situation where a method is being called and it is indeterminate if the sharedStore has already been set. My plan was to pop if an object existed or create a new object and re/push at the end, but it appears that a SharedStoreException is thrown if nothing is there. It would be nice to simple ask if the cache has already been set.

paterczm:
You mean it's indeterminate if the object in sharedStore was already set? Set parallel=false on the facaded api to ensure this is the case. Parallel=true processing makes sense only when api does not relay on the sharedStore.

dcrissman:
parallel is set to false, that isn't the issue. The issue is I have one facade method that calls another facade method. So the second method wouldn't know if the sharedStore has already been set or not. This new method would provide a means for the second method to check if a cache already exist as opposed to overriding it.

dcrissman:
@paterczm - What is your opposition to having a method to test if the cache contains something or not?

dcrissman:
Per discussion offline, will close this PR and go a different direction.

PR #333, opened 2016-04-18
remove seemingly do nothing method

dcrissman:
Does this method have a purpose? It looks like if just calls itself and is an infinite loop trap.


paterczm:
> Does this method have a purpose? It looks like if just calls itself and is an infinite loop trap.

It's an artifact created when copying logic from the deprecated DAOFacadeBase. It's not used anywhere.

PR #336, opened 2016-04-26
Healthchecks

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/ThreadMonitor.java
```diff
@@ -0,0 +1,279 @@
...
+                    } else {
+                        status=threadMap.get(thread);
+                    }
+                }
+            }
+            status.lastPingMsg=msg;
+            LOGGER.debug("Status:{}",status);
+            long now=System.currentTimeMillis();
+            if(status.status==Status.killed) {
+                ((Thread)thread).interrupt();
```
dcrissman: Could just use currentThread here to eliminate some casting.

migrator/src/main/java/com/redhat/lightblue/migrator/ThreadMonitor.java
```diff
@@ -0,0 +1,279 @@
...
+        LOGGER.debug("Ping from {}: {}",currentThread,msg);
+        if(currentThread instanceof MonitoredThread) {
+            MonitoredThread thread=(MonitoredThread)currentThread;
+            ThreadStatus status=threadMap.get(thread);
+            if(status==null) {
+                synchronized(this) {
+                    if(!threadMap.containsKey(thread)) {
+                        threadMap.put(thread,status=new ThreadStatus(thread));
+                    } else {
+                        status=threadMap.get(thread);
```
dcrissman: How do you know that status is not still null?

migrator/src/main/java/com/redhat/lightblue/migrator/ThreadMonitor.java
bserdar: Done

PR #345, opened 2016-05-03
Count threads correctly; when they end, remove the status record

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/ThreadMonitor.java
```diff
@@ -229,7 +240,7 @@ public void run() {
                     runNow=false;
                 } else {
                     try {
-                        waiter.wait(threadTimeout);
+                        waiter.wait(60l*1000l);
```
BVulaj: Put this in a var somewhere maybe.

bserdar: Done.

PR #350, opened 2016-05-17
Thread pool for lightblue calls

paterczm:
When lightblue does not respond for certain amount of time, facade walks away from the thread making the call. However, the thread is still alive in the background and can't be interrupted. The thread pool allows to limit the overall number of threads. The reason for this change is client resource protection.


facade/README.md
```diff
@@ -33,6 +33,13 @@ com.redhat.lightblue.migrator.facade.timeout.CountryDAO=2000
 means that facade will wait up to 2s for getCountry method on destination (Lightblue) service to finish execution and 2s for any other method. Timeouts make sense only when
 source service is being called in addition to the destination service, so that source service response can be returned when destination takes too long to respond.
 
+### Thread pool configuration
+
+Each call to lightblue is made in a separate thread. Those threads are managed using a singleton ThreadPoolExecutor instance (used by all facades using the same class loader). The pool is bounded to protect client's resources. By default it's 50 threads, but you can change it:
+```
+com.redhat.lightblue.migrator.facade.CountryDAO.threadPool.size=100
```
jewzaam: Is there a specific file this (and other) configuration has to be put into?  Isn't clear from the readme.

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -94,6 +105,21 @@ public ServiceFacade(D legacySvc, D lightblueSvc, String implementationName, Pro
 
         timeoutConfiguration = new TimeoutConfiguration(DEFAULT_TIMEOUT_MS, implementationName, this.properties);
 
+        // create ThreadPoolExecutor for all facades
+        synchronized("executor") {
```
jewzaam: Unless the string "executor" is a single instance shared this is not going to correctly protect this block of code.  Pick instead some static constant on this class, such as CONFIG_PREFIX, to synchronize on.

jewzaam: Actually, what about doing a check if executor is null first?  This way you don't block on synchronization if it is is already set.

```
if (executor == null) {
   synchronized (CONFIG_PREFIX) {
      if (executor == null) {
         ...
      }
   }
}
```

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -94,6 +105,21 @@ public ServiceFacade(D legacySvc, D lightblueSvc, String implementationName, Pro
...
+            if (executor == null) {
+                int threadPoolSize = Integer.parseInt(this.properties.getProperty(CONFIG_PREFIX+implementationName+".threadPool.size", "50"));
+
+                // cached means automatic thread reclamation
+                ThreadPoolExecutor threadPoolExecutor = (ThreadPoolExecutor) Executors.newCachedThreadPool();
+                // fixed size, no queuing
+                threadPoolExecutor.setCorePoolSize(threadPoolSize);
+                threadPoolExecutor.setMaximumPoolSize(threadPoolSize);
+                executor = MoreExecutors.listeningDecorator(threadPoolExecutor);
+                log.info("Initialized facade threadPool.size="+threadPoolSize);
```
jewzaam: Add implementationName to info log so it's clear what is being initialized

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
bserdar: For that to work, declare executor as volatile.

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -94,6 +105,21 @@ public ServiceFacade(D legacySvc, D lightblueSvc, String implementationName, Pro
...
+        synchronized("executor") {
+            if (executor == null) {
+                int threadPoolSize = Integer.parseInt(this.properties.getProperty(CONFIG_PREFIX+implementationName+".threadPool.size", "50"));
+
+                // cached means automatic thread reclamation
+                ThreadPoolExecutor threadPoolExecutor = (ThreadPoolExecutor) Executors.newCachedThreadPool();
+                // fixed size, no queuing
+                threadPoolExecutor.setCorePoolSize(threadPoolSize);
+                threadPoolExecutor.setMaximumPoolSize(threadPoolSize);
+                executor = MoreExecutors.listeningDecorator(threadPoolExecutor);
```
jewzaam: For the null checks around synchronization to work correctly save the created executor into a local variable first then assign to `executor`.  Used to be (aka maybe not a problem anymore) null check could see a field as not null while the assigned object was still being constructed.

bserdar: Still needs to be declared volatile

facade/README.md
paterczm: There isn't. You just pass properties object to Facade constructor.

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
paterczm: > Unless the string "executor" is a single instance shared this is not going to correctly protect this block of code. Pick instead some static constant on this class, such as CONFIG_PREFIX, to synchronize on.

It is a single instance, no? The simple syntax (no new keyword) will always use the same object. Example:
```
System.out.println("foo" == "foo");
System.out.println(new String("foo") == new String("foo"));

true
false
```

I'll add that additional null check.

paterczm: > For that to work, declare executor as volatile.

Good point. Never used that keyword before...

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
paterczm: You can have multiple facaded services in a single deployment, but only one pool, so it's not there is no 1 to 1 assoication between the pool and implementationName. But perhaps it's still better than nothing.

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -357,4 +397,19 @@ public TimeoutConfiguration getTimeoutConfiguration() {
     public void setTimeoutConfiguration(TimeoutConfiguration timeoutConfiguration) {
         this.timeoutConfiguration = timeoutConfiguration;
     }
+
+    /**
+     * Shutdown thread pool executor for all facades.
+     *
+     */
+    public static void shutdown() {
```
jewzaam: Is there any point to a shutdown?  When would this be used, except unit tests maybe?  It would need to be triggered by classloader teardown otherwise the executor will always exist.

paterczm: It's here for the sake of completeness. Could be called on service undeployment to avoid leaks during hot deployment.

jewzaam: But it isn't.
http://stackoverflow.com/questions/7602706/ejb-application-shutdown-hook

paterczm: Not sure what you mean. According to that thread, FacadeService.shutdown() could be called on singleton bean's pre destroy to cleanup the pool on undeployment.

jewzaam: Guess in a quick glance it looks like annotation @PreDestroy would be
needed for that to happen.  I don't see how declaring a static method on a
pojo will result in the EJB calling that method on shutdown.

paterczm: True. But you still need a method to call. ServiceFacade.shutdown() is that method.

jewzaam: Where is that method call?

paterczm: It should be implemented on the services side. Not part of this PR. Facade does not handle it's own lifecycle, because it doesn't know how it's being run (e.g. if it's EJB container).

paterczm: Are you suggesting to add EJB Singleton to manage Facade lifecycle to this project? It would require EJB deps.

paterczm: I'm not sure if we need to worry about this, since we restart app server after every change. It's not that much of a problem if there is a leak during hot deployment and we know how to fix it (a singleton to shutdown facade on pre-destroy) if we decide to do so.

jewzaam: We don't restart in all envs after update, but your point about this not
knowing its lifecycle covers it.  The EJB would need updated to hook into
this.  Thanks, knew I was missing something.

PR #356, opened 2016-05-20
Migration monitor

dcrissman:
Create nagios alert to check if there are MigrationConfigurations for which new jobs are not being created.

### DIFF
--- a/lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/Monitor.java
+++ b/lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/Monitor.java
@@ -1,12 +1,12 @@
 package com.redhat.lightblue.migrator.monitor;
 
+import java.util.ArrayList;
 import java.util.Date;
+import java.util.List;
 
 import org.joda.time.Period;
 import org.joda.time.format.PeriodFormat;
 import org.joda.time.format.PeriodFormatter;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.redhat.lightblue.client.LightblueClient;
 import com.redhat.lightblue.client.LightblueException;
@@ -19,13 +19,11 @@ import com.redhat.lightblue.client.response.LightblueDataResponse;
 import com.redhat.lightblue.migrator.MigrationConfiguration;
 import com.redhat.lightblue.migrator.MigrationJob;
 
-public class Monitor extends Thread {
-
-    private static final Logger LOGGER = LoggerFactory.getLogger(Monitor.class);
+public class Monitor {
 
     private final LightblueClient lightblueClient;
 
-    public Monitor(MainConfiguration cfg) {
+    public Monitor(MonitorConfiguration cfg) {
         if (cfg.getClientConfig() != null) {
             lightblueClient = new LightblueHttpClient(cfg.getClientConfig());
         } else {
@@ -33,29 +31,8 @@ public class Monitor extends Thread {
         }
     }
 
-    @Override
-    public void run() {
-        LOGGER.debug("Starting migrator monitor");
-        boolean stop = false;
-
-        while (!isInterrupted() && !stop) {
-            try {
-                process();
-            } catch (Exception e) {
-                LOGGER.error("Error during configuration load:" + e);
-            }
-
-            if (!stop) {
-                try {
-                    Thread.sleep(30000);
-                } catch (InterruptedException e) {
-                    stop = true;
-                }
-            }
-        }
-    }
-
-    private void process() throws Exception {
+    public void runCheck(final Notifier... notifiers) throws LightblueException {
+        List<String> configurationsMissingJobs = new ArrayList<>();
         for (MigrationConfiguration cfg : findMigrationConfigurations()) {
             long period = parsePeriod(cfg.getPeriod());
             Date endDate = new Date();
@@ -63,7 +40,17 @@ public class Monitor extends Thread {
 
             int jobs = countMigrationJobs(cfg.getConfigurationName(), startDate, endDate);
             if (jobs <= 0) {
-                //TODO some sort of alert
+                configurationsMissingJobs.add(cfg.getConfigurationName());
+            }
+        }
+
+        if (configurationsMissingJobs.isEmpty()) {
+            for (Notifier n : notifiers) {
+                n.sendSuccess();
+            }
+        } else {
+            for (Notifier n : notifiers) {
+                n.sendFailure(configurationsMissingJobs);
             }
         }
     }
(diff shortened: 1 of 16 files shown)


lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/Monitor.java
```diff
@@ -0,0 +1,93 @@
...
+                Projection.includeField("period"));
+
+        return lightblueClient.data(findConfigurations, MigrationConfiguration[].class);
+    }
+
+    private int countMigrationJobs(String configurationName, Date startDate, Date endDate) throws LightblueException {
+        DataFindRequest findJobs = new DataFindRequest(MigrationJob.ENTITY_NAME);
+        findJobs.where(
+            Query.and(
+                Query.withValue("configurationName", Query.eq, configurationName),
```
bserdar: instead of scheduled date, use consistencyChecker,jobRangeBegin and jobRangeEnd

dcrissman: consistencyChecker.jobRangeBegin is a string type?

bserdar: It is string. It can be an Id, date, etc.

dcrissman: My assumption is that at least one migrationJob will be created for every period on a MigrationConfiguration (that has a period set). I am not understanding why jobRangeBegin is better than scheduledDate for this?

bserdar: The job range is the range for which the job is created. ScheduledDate is
simply the date it is scheduled. There should be one job for each job range.

dcrissman: Then is period the wrong field to be looking at on migrationConfiguration? I can't take a date calculated from a period and compare it to a string field that could contain a non-date value?

bserdar: You can take a date, convert it to string, and compare it to the string
field.

dcrissman: I hear what you are saying about wanting to use the the jobRange fields, but I still have some concerns about it.

Firstly, that would then be doing a string compare instead of a date compare, which could produce different results. 
Secondly, there is no way to guarantee the date format used in the string field. 
Thirdly, there is no way to know for sure that the string field even contains a date and not some other sort of value. You said yourself it could be an id field.

bserdar: Your concerns are valid. This can only be used if the ranges use a know
date syntax, and that those ranges are really dates. But what is the
alternative? Those fields are the only ones that tell what range the job is
supposed to process.

jewzaam: The point is to monitor for the creation of jobs, not that there's a contiguous range of jobs being created.  That may be a separate issue to tackle.  Could we use the scheduled date to just check that we're getting at least something generated for each configured and active migration?

bserdar: We can use scheduled date as an indication that the consistency checker is
still alive. There should be new jobs for every "period" with scheduled
date close to what the period should be.

bserdar: Also, keep in mind that the consistency checker follows the schedule two
periods behind. Maybe we should fix that as well, I meant it to follow one
period behind, somehow it turned out this way.

dcrissman: I had actually thought about the monitor possibly checking a period or two behind just to provide some slack. This would be fine for jobs that generate every hour, but dailies would become problematic.

dcrissman: @bserdar - so are there any changes required then?

jewzaam: Could it create future dated jobs with a future scheduled date instead of
being behind?  A time X create job for X + period?

bserdar: It could, but why?

lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/Monitor.java
```diff
@@ -0,0 +1,93 @@
...
+            lightblueClient = new LightblueHttpClient();
+        }
+    }
+
+    public void runCheck(final Notifier... notifiers) throws LightblueException {
+        List<String> configurationsMissingJobs = new ArrayList<>();
+        for (MigrationConfiguration cfg : findMigrationConfigurations()) {
+            long period = parsePeriod(cfg.getPeriod());
+            Date endDate = new Date();
+            Date startDate = new Date(endDate.getTime() - period);
```
bserdar: use something like ... - 3*period, or 2*period

lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/Monitor.java
jewzaam: You mentioned they were behind.  Why not instead be ahead?

bserdar: Possible. So instead of creating jobs for the past period, create jons for
the coming one.

Can you add a ticket to migrator?

jewzaam: https://github.com/lightblue-platform/lightblue-migrator/issues/358

lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/Monitor.java
dcrissman: I assume you mean 2 or 3 periods ago? If so, fine with me, but what if the period is a day? Then the check wouldn't be for 2 or 3 days ago... are we ok with that?

dcrissman: I gave the periods a multiplier and made it configurable.

dcrissman:
@bserdar, @jewzaam - I think all the requested changes have been made.

bserdar:
+1

PR #359, opened 2016-05-25
Remove deprecated facade

paterczm:
Since we plan to push all facaded services soon, we won't be needing that anymore.


dcrissman:
Happily

PR #360, opened 2016-05-26
Interrupt correctly

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/ConsistencyCheckerController.java
```diff
@@ -206,109 +206,103 @@ private boolean migrationJobsExist() {
...
-                        LOGGER.info("There are migration jobs for {}, not running consistency checker this time",
+        LOGGER.debug("Starting consistency checker controller for {} with period {}",migrationConfiguration.getConfigurationName(),
+                     migrationConfiguration.getPeriod());
+        long period=parsePeriod(migrationConfiguration.getPeriod());
+        while(!stopped) {
+            Breakpoint.checkpoint("CCC:start");
+            LOGGER.debug("Consistency checker {} woke up",migrationConfiguration.getConfigurationName());
+            // Lets update our configuration first
+            
+            MigrationConfiguration newCfg=migrationConfiguration;
```
dcrissman: Does this need to be set with each iteration?

migrator/src/main/java/com/redhat/lightblue/migrator/AbstractController.java
```diff
@@ -27,16 +27,17 @@
...
-    public AbstractController(Controller controller, MigrationConfiguration migrationConfiguration, String threadGroupName) {
-        this.migrationConfiguration = migrationConfiguration;
-        this.controller = controller;
-        lbClient = controller.getLightblueClient();
-        locking = lbClient.getLocking("migration");
-        if (migrationConfiguration.getMigratorClass() == null) {
-            migratorClass = DefaultMigrator.class;
-        } else {
+    protected final HashSet<String> myLocks=new HashSet<>();
+    protected boolean stopped=false;
```
dcrissman: Is stopped read in this class? If not, does it maybe belong on a lower level?

migrator/src/main/java/com/redhat/lightblue/migrator/AbstractController.java
```diff
@@ -47,6 +48,11 @@ public AbstractController(Controller controller, MigrationConfiguration migratio
         migratorThreads = new ThreadGroup(threadGroupName);
     }
 
+    public void setStopped() {
```
dcrissman: just stop()?

migrator/src/main/java/com/redhat/lightblue/migrator/ConsistencyCheckerController.java
bserdar: It runs once every period, and reloads the configuration to see if it should still run.

migrator/src/main/java/com/redhat/lightblue/migrator/AbstractController.java
bserdar: It is not read here, but when the thread needs to stop, it is set at this level. All controller classes inherit AbstractController.

migrator/src/main/java/com/redhat/lightblue/migrator/AbstractController.java
bserdar: Clashes with Thread.stop()

PR #376, opened 2016-06-09
Create high inconsistency rate monitor

dcrissman:


lightblue-migration-monitor/src/main/java/com/redhat/lightblue/migrator/monitor/HIR/HIRMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+        for (MigrationJob job : jobs) {
+            JobExecution execution = job.getJobExecutions().get(0);
+
+            processedDocumentCount += execution.getProcessedDocumentCount();
+            overwrittenDocumentCount += execution.getOverwrittenDocumentCount();
+        }
+
+        int percent = 0;
+        if(processedDocumentCount != 0){
+            percent = 100 * (overwrittenDocumentCount / processedDocumentCount);
```
bserdar: Did you really test this? This is int division,and won't return a meaningful percent. Either multiply with 100 first, then divide, or convert everything to double.

dcrissman: Not well enough. I changed the testRunCheck_ThresholdExceeded to have a result with an actual decimal point.

PR #377, opened 2016-06-14
Controller thread dies somehow, prevent that

bserdar:


migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
```diff
@@ -28,7 +28,7 @@
     private final LightblueClient lightblueClient;
     private final Map<String, MigrationProcess> migrationMap = new HashMap<>();
     private final ThreadMonitor threadMonitor;
-    private boolean stopped=false;
+    private volatile boolean stopped=false;
```
dcrissman: I know what volatile does, but what is the reason for using it here?

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
```diff
@@ -224,15 +224,10 @@ public void run() {
                 MigrationConfiguration[] cfg=getMigrationConfigurations();
                 createControllers(cfg);
                 Breakpoint.checkpoint("Controller:createconfig");
-            } catch (Exception e) {
+                Thread.sleep(30000);
+            } catch (Throwable e) {
```
dcrissman: Why catch the InterruptedException here? There is no need to log it as it simply means the thread has been terminated.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
bserdar: Agreed, no need.

migrator/src/main/java/com/redhat/lightblue/migrator/Controller.java
bserdar: This thread somehow terminates. Any exception thrown here, including InterruptedException, doesn't mean the thread should terminate. It should only terminate when stopped is set.

PR #379, opened 2016-06-21
Try to interrupt timed out calls

paterczm:
Restoring interrupting timed out lightblue calls.

This should help with "java.lang.OutOfMemoryError: Unable to create new native thread" errors, although it's not full protection, like thread pools.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
@@ -295,7 +295,7 @@ public class ServiceFacade<D extends SharedStoreSetter> implements SharedStoreSe
                 }
             } catch (TimeoutException te) {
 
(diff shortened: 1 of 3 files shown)


facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -294,6 +294,16 @@ private boolean shouldCheckConsistencyConsistency(FacadeOperation facadeOperatio
                     lightblueEntity = getWithTimeout(listenableFuture, methodName, facadeOperation, shouldSource(facadeOperation));
                 }
             } catch (TimeoutException te) {
+
```
bserdar: We should not interrupt write calls, we might interrupt in the middle of  related write operations, leaving corrupt data.

paterczm: Good point.

PR #383, opened 2016-07-05
Trying to fix intermittent facade test failures

paterczm:
For some reason java8 builds take much longer than java7 builds (e.g. https://travis-ci.org/lightblue-platform/lightblue-migrator/builds/142510528). This could be responsible for test failures.


bserdar:
How about using Breakpoint? it is in migrator package.

paterczm:
The java7 build is much faster because it doesn't run tests at all.

dcrissman:
> The java7 build is much faster because it doesn't run tests at all.

That is correct. We do this because the upstream integration tests require jdk8.

paterczm:
Took a lot of try and error in travis, but I think I finally got it.

PR #388, opened 2016-08-10
Select job type (generated or not) according to weights specified in configuration

paterczm:
2 things changed:
* Consistency checker wakes up to generate new jobs regardless of available migration jobs.
* Migrator retrieves either generated or non-generated jobs, according to weights specified in configuration. Migration and consistency checking can run in parallel and their participation can be controlled.

This is not very clean approach, but at least I didn't need to touch the thread management logic, which I find really complex.

TODO: unit test

### DIFF
--- a/migrator/pom.xml
+++ b/migrator/pom.xml
@@ -74,12 +74,12 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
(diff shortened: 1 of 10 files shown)


bserdar:
There is a migrationJob  index on  (configurationName,status,scheduledDate), lets change that to (configurationName,status,scheduledDate,generated)

PR #392, opened 2016-08-22
Logging slow calls also when there is no timeout

paterczm:
Facade will not timeout Lightblue call until source responds successfully. This change logs slow call warning independently of source and can help identify problems in Lightblue when source takes very long to respond or throws an exception.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/TimeoutConfiguration.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/TimeoutConfiguration.java
@@ -43,6 +43,7 @@ public class TimeoutConfiguration {
     public static final String CONFIG_PREFIX = "com.redhat.lightblue.migrator.facade.";
 
     private long defaultTimeoutMS;
+    private long defaultSlowwarnMS;
     private String beanName;
     private Properties properties;
     private boolean interruptOnTimeout = true;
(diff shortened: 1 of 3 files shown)


facade/src/main/java/com/redhat/lightblue/migrator/facade/TimeoutConfiguration.java
```diff
@@ -118,7 +118,17 @@ public long getMS(String methodName, FacadeOperation op, Type type) {
                 log.debug("{} config not found for bean {} using global timeout", type, beanName);
             }
 
-            timeout = defaultTimeoutMS;
+            switch (type) {
+                case timeout: {
+                    timeout = defaultTimeoutMS; break;
+                }
+                case slowwarning: {
+                    timeout = 2 * defaultTimeoutMS; break;
```
dcrissman: I am generally good with this change, but am not sure that I understand the point of the Type enum. I see that it previously existed, but until this it doesn't look like it is used for anything except logging. Why not just let the constructor handle this?

paterczm: Yes, it's used only for logging. I'm not sure what you mean by 
> let the constructor handle this?

Different types of timeouts may have different values. There is only one TimeoutConfiguration instance per facade.

dcrissman: >> let the constructor handle this?

> Different types of timeouts may have different values. There is only one TimeoutConfiguration instance per facade

I don't understand the need to double the defaultTimeoutMS when it is set by the constructor.

Maybe just some documentation that states that the slowwarning type will just double the defaultTimeoutMS.

paterczm: I see what you mean. Moved that to the constructor and changed so that defaultSlowwarnMS = defaultTimeoutMS. I didn't want both timeout and slow call warns to be logged for the same calls, but I realized this can be confusing.

dcrissman:
Sorry, hit the wrong button there. Merged!

PR #395, opened 2016-08-30
Fail for non-annotated methods in proxy phase

paterczm:
When a method is not annotated, facade sends it to source/legacy bean. This is not a correct behavior in the proxy phase, when the source/legacy bean should not be used anymore. Throw an exception instead.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/proxy/FacadeProxyFactory.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/proxy/FacadeProxyFactory.java
@@ -15,6 +15,7 @@ import org.slf4j.LoggerFactory;
 import com.redhat.lightblue.migrator.facade.ServiceFacade;
 import com.redhat.lightblue.migrator.facade.ServiceFacade.FacadeOperation;
 import com.redhat.lightblue.migrator.facade.sharedstore.SharedStoreSetter;
+import com.redhat.lightblue.migrator.features.LightblueMigration;
 
 /**
  * Creates a dynamic proxy implementing given interface. The calls to the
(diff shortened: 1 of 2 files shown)


alechenninger:
LGTM

facade/src/main/java/com/redhat/lightblue/migrator/facade/proxy/FacadeProxyFactory.java
```diff
@@ -100,12 +101,12 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
 
                 switch (t) {
                     case LEGACY: {
-                        log.debug("Not a facade operation, proxy passing to legacy");
+                        log.debug("Method \"{}\" is explicitly annotated to call legacy", method);
```
BVulaj: Is debug okay for this kind of information?  Do we have debug enabled in dev?  It may be hard to track these down / monitor them.

facade/src/main/java/com/redhat/lightblue/migrator/facade/proxy/FacadeProxyFactory.java
```diff
@@ -124,10 +125,16 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 return svcFacade.callSvcMethod(FacadeOperation.WRITE, wo.parallel(), method, args);
             }
 
-            log.debug("Not a facade operation, proxy passing to legacy");
+            if (!LightblueMigration.shouldReadSourceEntity() && !LightblueMigration.shouldWriteSourceEntity()) {
+                // this method is not annotated and legacy/source is no more
+                throw new IllegalStateException("Method \""+method+"\" is not annotated for facade and legacy source is disabled!");
+            } else {
+                // this can cause problems when entering proxy phase
+                log.warn("Method \"{}\" is not annotated for facade. Proxy passing to legacy.", method);
```
BVulaj: Maybe ignore my initial comment. This is probably okay.

facade/src/main/java/com/redhat/lightblue/migrator/facade/proxy/FacadeProxyFactory.java
paterczm: Those are explicitly annotated methods e.g. ```@Direct(target = LEGACY)```. In some cases we may not migrate entire service at once. We can grep for those in code.

PR #396, opened 2016-09-15
Kinda proxy phase support

paterczm:
Kinda proxy phase means:
1. writes go to both legacy and Lightblue services,
2. if write fails in Lightblue, exception is thrown (in dual read phase it is swallowed!),
3. if write fails in legacy, exception is swallowed (in dual read phase it is thrown!),
4. reads go only to Lightblue service,
5. consistency checks are disabled,
6. all Lightblue timeouts are disabled.

For write operations, legacy is still going to be called first to ensure all generated data (ids, timestamps, etc.) can be passed to shared store. Legacy is still the source of this shared information.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
@@ -208,7 +208,7 @@ public class ServiceFacade<D extends SharedStoreSetter> implements SharedStoreSe
         }
     }
 
-    private boolean shouldCheckConsistencyConsistency(FacadeOperation facadeOperation) {
+    private boolean shouldCheckConsistency(FacadeOperation facadeOperation) {
         switch (facadeOperation) {
             case READ:
                 return LightblueMigration.shouldCheckReadConsistency();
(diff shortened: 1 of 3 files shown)


facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -275,7 +275,13 @@ private boolean shouldCheckConsistencyConsistency(FacadeOperation facadeOperatio
             try {
                 legacyEntity = (T) method.invoke(legacySvc, values);
             } catch (InvocationTargetException e) {
-                throw e.getCause();
+                if (shouldDestination(facadeOperation) && !shouldCheckConsistency(facadeOperation)) {
+                    // Lightblue is going to be called and consistency checker is disabled
+                    // swallow the exception from legacy
+                    log.warn("Legacy call " + implementationName + "." + callStringifier + " threw an exception. Returning data from Lightblue.", e.getCause());
+                } else {
+                    throw e.getCause();
```
alechenninger: Should we cancel the lightblue call future if it is in progress at this point? That being said I realize this is same as previous behavior.

alechenninger:
LGTM! One comment, but this is about behavior that already existed. Let me know if you want to merge now, or you can merge at will.

facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
paterczm: It's a good point, let's handle this in a separate pr.

alechenninger: K, sounds good

PR #399, opened 2016-09-15
Try to cancel lightblue call if source call failed

paterczm:
Using isInterruptOnTimeout configuration key for interrupt on source failure as well.

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
@@ -219,6 +219,16 @@ public class ServiceFacade<D extends SharedStoreSetter> implements SharedStoreSe
         }
     }
 
+    private void cancel(ListenableFuture listenableFuture) {
+        if (listenableFuture != null) {
+            try {
+                listenableFuture.cancel(true);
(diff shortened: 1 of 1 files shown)


facade/src/main/java/com/redhat/lightblue/migrator/facade/ServiceFacade.java
```diff
@@ -219,6 +219,16 @@ private boolean shouldCheckConsistency(FacadeOperation facadeOperation) {
         }
     }
 
+    private void cancel(ListenableFuture listenableFuture) {
+        if (listenableFuture != null) {
+            try {
+                listenableFuture.cancel(true);
+            } catch (Exception e) {
+                log.error("Failed to cancel lightblue call", e);
```
alechenninger: May want to log if cancel returns false similarly

paterczm: You mean a debug statement that it couldn't be cancelled? I don't think it calls for a warning, error even more so.

alechenninger: Maybe a warning, but I agree not an error in that case. Up to you.

PR #407, opened 2017-03-21
Not using JSON compare for large result sets b/c it consumes a lot of memory

paterczm:
Service facade uses 2 libraries to diff api responses: JSONCompare and Jiff. Jiff is faster and consumes less memory, but provides less details (stops on first diff encountered and does not describe array element differences). This change will use JSONCompare to report for all responses under 65kB and Jiff for bigger ones (configurable).

### DIFF
--- a/facade/src/main/java/com/redhat/lightblue/migrator/facade/ConsistencyChecker.java
+++ b/facade/src/main/java/com/redhat/lightblue/migrator/facade/ConsistencyChecker.java
@@ -2,6 +2,7 @@ package com.redhat.lightblue.migrator.facade;
 
 import java.io.IOException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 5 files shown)


bserdar:
Instead of comparing lenghts, why not use jiff to compare, and then if docs are different, use jsoncompare to get the differences? I think this is the way it is done in migrators.

paterczm:
> Instead of comparing lenghts, why not use jiff to compare, and then if docs are different, use jsoncompare to get the differences? I think this is the way it is done in migrators.

Already doing that, but still running out of memory because of huge inconsistencies in product result sets.

paterczm:
The logic is following:
1. Use jiff to compare responses. If equal, return.
2. If different, check size of responses.
3. If size < 64kB, use JSONCompare to produce inconsistency (diff) for analysis.
4. If size >= 64kB, use Jiff to produce inconsistency (diff) for analysis.

bserdar:
Yes, saw that

